PR #29, opened 2016-03-22
Fedger.io

brakmic:
This is a publicly available API comprising of several sub-APIs: company, fundings, geodata, news etc. 
The API itself is proprietary as the company behind it charges for its usage but there's a certain amount of "free calls" for every registered consumer.
Registering the API key is very easy and for free. I wrote a PureScript library to access this API. The backend logic is in JS, so it can be easily used without PureScript as well.

Not sure if such APIs are welcome here, so feel free to reject this PR.


toddmotto:
Wow, what an amazing API doc, website is stunning. Let's add it in, some other APIs have limits to them, GitHub/etc.

brakmic:
Thanks! :+1:

PR #53, opened 2016-03-23
Both BreweryDB and LCBI needs API-KEY (registered users)

juanbrujo:


toddmotto:
I think a lot of these require license keys to use, any thoughts to a new column in the tables?

juanbrujo:
Yeah, maybe a new column:

| API | Description | OAuth | API-Key | Link |
|---|---|---|---|---|
| BreweryDB | Beer | No | Yes | [Go!](http://www.brewerydb.com/developers) |
| LCBO API | Alcohol | No | Yes | [Go!](https://lcboapi.com/) |

PR #64, opened 2016-03-24
Add exchange rates api

unseen1980:
add exchange rates api + a-z sort order


toddmotto:
Awesome, I was going to clean this up once the PRs slowed down a little, thanks for diving in! :)

PR #73, opened 2016-03-25
Update README.md

prabhakar267:


wesbos:
This needs a better description - looks like a nice API though

PR #74, opened 2016-03-25
Add jokes section and include Chuck Norris API

nigelhanlon:
Add a joke section for API's that provide jokes. I have included the Chuck Norris API to get things started.


wesbos:
Already present under Media - thanks though!

PR #79, opened 2016-03-26
Corrected AniList OAuth support

joshstar:
The AniList API does implement OAuth based authentication, for some reason it wasn't mentioned any where in the documentation but thats [been amended now](http://anilist-api.readthedocs.org/en/latest/authentication.html).

Thanks


toddmotto:
Awesome, thanks!

PR #101, opened 2016-03-28
add nyc

abhisharma2:


YannBertrand:
It doesn't seem to use OAuth. It should be good to go if you change this column to `No, but an api key is mandatory`

PR #109, opened 2016-03-29
Add Nutritionix

johnstew:
Added in Nutritionix. Not sure exactly what the category for this should be. I left it as **Health** for now but was also thinking about just using **Food**. 

Either way Nutritionix is an API I used recently for searching for nutrition facts about all kinds of food and it was pretty good. Think its a good addition to the list.


wesbos:
sweet - thanks!

PR #112, opened 2016-03-29
Add Zip Codes

mayra-cabrera:
Add a zip codes section for API's that provide this kind of information. I have included the Mexico API to get things started.

Think its a good addition to the list.


mikestreety:
Moved to Geocoding section with other post codes

PR #120, opened 2016-03-30
Add Amadeus Travel Innovation Sandbox

tadhgpearson:
Add sandbox.amadeus.com to the Transportation category


tadhgpearson:
I thought this would be a good API to add. You can do flight, car, hotel, and rail searches, and it includes a bunch of support APIs too. However, it is limited to non-commercial use, and usage quotas apply.

PR #121, opened 2016-03-30
Added public APIs from apilayer.com

apilayer:
Hi there!

We noticed that one of our APIs was added by a member, so we went ahead and added a few more publicly available apilayer APIs. 

Thanks,
apilayer Team


YannBertrand:
Wow, a lot of new APIs here :smile:!

All of those look pretty interesting. What's up with giflayer though?

dominikkukacka:
We discontinued giflayer beacause of low interest rate :)

PR #141, opened 2016-04-01
Added a recipe puppy api to the list

prashant-andani:
A recipe finder api


wesbos:
Ohh this is a really good one! Thanks

PR #142, opened 2016-04-01
Add Star Wars API

emaraschio:


crocodele:
It's already present, but _Movies_ makes more sense than _Games & Comics_ IMHO.

wesbos:
Might rename that title - closing this as a dupe though. Thanks!

PR #168, opened 2016-04-19
Adding Holiday API

joaopedronardari:


wesbos:
Looks like this one will need an API key pretty soon, so an update would be appreciated!

PR #200, opened 2016-07-03
Fix link to Spotify's Web API

thelinmichael:
Replaced link to Spotify's Web API with one that points to the Web API's landing page. (The old link pointed to a migration page from the old Web API.)


jbrooksuk:
Thanks!

PR #207, opened 2016-08-01
Adding the Nomad List API

swbloom:
The documentation isn't great, but the endpoint is very cool:

From https://nomadlist.com/faq:
"Nomad List finds you the best places in the world to live and work remotely.	It collects over 50,000 data points about 500+ cities around the world, from cost of living, temperature to safety. "


YannBertrand:
😀

PR #208, opened 2016-08-06
Added wger workout manager API

emlagowski:
Added wger API. Useful for fitness apps. Contains public and private endpoints. All informations is here https://wger.de/en/software/api


toddmotto:
Thanks!

PR #216, opened 2016-09-02
Adds an anti-malware section.

iangcarroll:
Adds an anti-malware section.

I founded Certly, so there's some bias here, but I added all the other (free) ones that I'm aware of.


jbrooksuk:
Thank you! :)

PR #221, opened 2016-09-06
Editing all Mashape apis

austinhuang0131:
...since they do have a central key for all these APIs that are listed in Mashape marketplace.

_Also, if you want (mandatory), I suggest you give me write access since I may add things frequently._


YannBertrand:
Thanks :)

jbrooksuk:
> Also, if you want (mandatory), I suggest you give me write access since I may add things frequently.

Don't worry, we're all around to merge.

PR #229, opened 2016-10-12
Added Forismatic API 

patrickk93:
Added Forismatic API, which provides "the most inspiring expressions of mankind." Added under the 'Media' section. 

Also edited Faceplusplus link to match other links in the list (GO! -> Go!)


toddmotto:
Thank you! :)

PR #235, opened 2016-10-24
Updated OpenWeatherMap OAuth - http://openweathermap.org/appid

d-scott:
Added 'apiKey' caveat to the OpenWeatherMap OAuth column: http://openweathermap.org/appid


davemachado:
+1

PR #237, opened 2016-10-29
Move Planets section into Science

davemachado:
Similar to the Media section, by moving the Planets section into Science, it creates a more concise way for users to search (i.e. no need to check Science, and then go back and look for another section to cover what they are looking for).


mikestreety:
Good idea dave - merged

PR #239, opened 2016-11-07
Remove Stattleship API

xerona:
Josh and David couldn't raise enough to keep the api public.


jbrooksuk:
Do you have a link to the email that Josh references [here](http://blog.stattleship.com/open-letter-to-our-api-community/)? He hasn't explicitly said the service is down.

xerona:
Josh sent an email the same day where he said it would be shut down saturday.
>Tomorrow is a real cut-off date. We hope you and others step up. We feel really good about providing this data and we don’t regret the time and money we’ve spent supporting this community. We are die hard sports fans and data junkies, too. 

I only have that and a demo app where api calls no longer work. https://xerona.github.io/sport-stats/
If I see more public information, I'll pass it along.

mikestreety:
It does indeed seem to now be a "paid-for" service - understandably so.

For non-commercial projects, you can choose to donate - but for anything commercial there is a [pricing structure](https://www.patreon.com/stattleship) in place.

I would say we should remove this - unless anything thinks otherwise?

mikestreety:
As there has been no objection in 20 days - i'm merging 😄

PR #251, opened 2016-12-10
Add index

siempredelao:
Added an index for easy access.


jbrooksuk:
Great idea! Thanks :)

Issue #203, opened 2016-07-12
Migrate to a website


yannbertrand:
The time has come. We now have a lot of public JSON APIs.

It's a pain to search the ones we're interested in.

What about moving the data all of us have collected to a website with a powerful search engine? (take a look at https://twofactorauth.org/ for a good exemple)

jbrooksuk:
I was thinking more along the lines of http://justdelete.me

YannBertrand:
Love it! :smile:

toddmotto:
Good idea, at the moment I don't have the time to do this though :/ how were you thinking of doing it, using Jekyll to render out all APIs from `_posts` data or something?

RyanNerd:
May I make a suggestion for moving this to a website?
Use [Elastic search](https://www.elastic.co/downloads/elasticsearch) as your search engine back-end.
I've been writing web apps in Ember and have been loving it.
A basic search web site shouldn't be too difficult to implement. I have a bit of time on my hands and would be willing to design a proof of concept if you are interested.
https://github.com/RyanNerd

YannBertrand:
I was thinking of migrating all the data to a json file and use it as a database. Do you think Jekyll would be an easier solution?

RyanNerd:
I'm not familiar with Jekyll so what I suggest below may be a bad idea, or a good idea:

I can provision a LAMP stack pretty quickly. I am using digitalocean.com as my host. For around $5/mo for a basic VM isn't a bad price. I use them to prototype web projects and I can scale my VM up easily if things get busy.
Writing the front-end will actually take me longer than setting up a back-end.
Let me know your thoughts.

davemachado:
Using a basic Digital Ocean VM is super easy to spin up and I've heard through the grapevine that sometimes DO will give you credit for a VM if the project is being hosted is popular enough/the README mentions the corresponding website is hosted by Digital Ocean (yay marketing!). Just some more food for thought to possibly make this task a bit financially easier.

RyanNerd:
DigitalOcean VM for a small droplet is around $5/mo. This is not a huge database and just needs a basic UI where this small VM would be just fine.

jbrooksuk:
Yeah, DO is awesome. They very kindly sponsor two of my projects. @RyanNerd is right in that it only needs a small VM 👍

davemachado:
I've reached out to someone on the Sales team at DO to see if they'd be willing to sponsor this undertaking and provide us with a shiny new VM.

[...]

PR #259, opened 2016-12-29
[WIP] JSON

jbrooksuk:
Working towards #203 

---

Firstly, we have to start the long and tedious process of converting the README.md file to a JSON list. I've completed two categories, but I wanted to get feedback on the format I've used.

I've also used `gulp` and `gulp-jsonlint` because this can become part of the CI process for new contributions.


jbrooksuk:
@yannbertrand probably 😄 it's more to discuss the structure before we generate it.

yannbertrand:
Yup! We should probably check out how github-pages work. I think we need to put that in a `gh-pages` branch if we decide to use it

jbrooksuk:
You can tell GitHub to use a specific branch or a `docs` directory.

yannbertrand:
Good to know 👍

The format seems fine to me 😉

mikestreety:
I would personally put the category as an attribute on each API (it could then expand to have many cateogories if required). Apart from that, it looks good.

jbrooksuk:
> I would personally put the category as an attribute on each API (it could then expand to have many cateogories if required). Apart from that, it looks good.

Perhaps an array of categories?

mikestreety:
Yeah, that's what I was thinking.

yannbertrand:
```
[
    {
        "name": "IUCN",
        "description": "IUCN Red List of Threatened Species",
        "auth": "token",
        "link": "http://apiv3.iucnredlist.org/api/v3/docs",
        "categories": ["Animals"]
    }, {
        "name": "Petfinder",
        "description": "Adoption",
        "auth": "no",
        "link": "https://www.petfinder.com/developers/api-docs/",
        "categories": ["Animals"]
    }, {
        "name": "RescueGroups",
        "description": "Adoption",
        "auth": "no",
        "link": "https://userguide.rescuegroups.org/display/APIDG/API+Developers+Guide+Home",
        "categories": ["Animals"]
    }, {
        "name": "AniList",
        "description": "AniList Anime API",
        "auth": "oAuth",
        "link": "http://anilist-api.readthedocs.org/en/latest/#",
        "categories": ["Anime"]
    }, {
        "name": "Hummingbird",
        "description": "Hummingbird Anime API",
        "auth": "no",
        "link": "https://hummingbird.me/",
        "categories": ["Anime"]
    }
]
```

jbrooksuk:
Better 👍

davemachado:
While we are laying the structure down for the JSON, should we also add the HTTPS-support as an attribute? Adding that specification is another open issue right now, so it might be good just to try to future-proof this structure as we go.

mikestreety:
I was just thinking about that. I agree.

jbrooksuk:
For sure!

davemachado:
Discussion for this carries on in #368

apis.json
```diff
@@ -0,0 +1,38 @@
...
+                "link": "https://www.petfinder.com/developers/api-docs/"
+            }, {
+                "name": "RescueGroups",
+                "description": "Adoption",
+                "auth": "no",
+                "link": "https://userguide.rescuegroups.org/display/APIDG/API+Developers+Guide+Home"
+            }
+        ]
+    }, {
+        "category": "Anime",
```
kuhnrl30: Wanted to point out that the JSON format used here is different than what was last recommended in #259: 
```
[
    {
        "name": "IUCN",
        "description": "IUCN Red List of Threatened Species",
        "auth": "token",
        "link": "http://apiv3.iucnredlist.org/api/v3/docs",
        "categories": ["Animals"]
    },
```
The difference being how the category is used.  The first format opens us to duplicate entries if an API is put in multiple categories creating a risk that they aren't identical.  i.e. the link gets updated in one entry but not the duplicate.  

Are we ok with that?  I just recently started following so I'm not sure what the protocol is for making decisions on this project.

davemachado:
As far I can tell, we work to limit entries to one category that best fits
the service offered. My opinion is to keep it that way so we don't end up
with a huge list of duplicate entries (for example, Instragram could end up
in Social, Photography, Development, etc since one could argue their API
offers some type of service for each of those categories. We can change the
structure of how entries are entered into this collection, but I fear the
list will become too cluttered if we allow duplicate entries.

PR #260, opened 2016-12-29
Hummingbird is dead, Long live Kitsu

Epictek:
Removed Hummingbird and updated with the Kitsu api details.

Source: https://medium.com/heykitsu/hummingbird-is-dead-long-live-kitsu-bda6ccfbbcce#.xixj9r30c
https://forums.hummingbird.me/t/kitsu-api-documentation/33367

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ For information on contributing to this project, please see the [contributing gu
 | API | Description | Auth | Link |
 |---|---|---|---|
 | AniList | AniList Anime API | `oAuth` | [Go!](http://anilist-api.readthedocs.org/en/latest/#) |
-| Kitsu | Kitsu Anime API | No | [Go!](http://docs.kitsu17.apiary.io/) |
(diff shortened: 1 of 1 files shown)


davemachado:
[Another source](https://forums.hummingbird.me/t/goodbye-hummingbird-hello-kitsu/33238) confirms Hummingbird is down and Kitsu, still unpaid, is taking its place.

Thanks @Epictek !

davemachado:
@Epictek it looks like the new API [uses OAuth for authentication](http://docs.kitsu17.apiary.io/#introduction/authentication).Can you please update the listing as needed?

Epictek:
Sorry the contribution guide seems kind of misleading as it said to put No if authentication is not required to use whereas the API can work without authentication (read-only)

davemachado:
valid point, @Epictek . In the case of this API, however, we'll want to cover what type of authentication is used when required. Thanks again for keeping us updated!

PR #262, opened 2017-01-13
Fix Riot Games

Gheoan:
https://developer.riotgames.com/docs/api-keys


davemachado:
Thanks for the updates @Gheoan !

PR #265, opened 2017-01-17
Add Foursquare API info

PaulCapestany:


davemachado:
Hi @PaulCapestany , thanks for the PR! I noticed you listed Foursquare under both Geocoding and Social. While this API does appear to have support for retrieving locations, the main focus of Foursquare is to provide a social platform between users. That being said, for that reason and to avoid service duplication, could you please remove the geocoding index? Thanks!

PaulCapestany:
@davemachado np, done! 

Since this is a popular repo with lots of contributors who may run into a similar categorization issue, I figured I'd mention that I didn't see any guidelines about how to handle potentially listing an API under multiple categories within [CONTRIBUTING.md](https://github.com/toddmotto/public-apis/blob/master/CONTRIBUTING.md). I'd assumed that it'd be ok based on https://github.com/toddmotto/public-apis/pull/259#issuecomment-269635584, so perhaps something worth clarifying somewhere?

davemachado:
@PaulCapestany This issue of duplicate listings is totally valid, and as such I've just updated the contributing guide to include a note regarding API's that can span multiple categories (PR #266). Thanks for bringing this up!

PR #271, opened 2017-01-26
Add Uber API

agraebe:
The Uber API is free to use


davemachado:
This presents an interesting situation. While the Uber API may be free to use, the entire Uber service is paid and does not offer any sort of "free tier." Currently, this list does not contain any paid solution APIs except for FitBit, which would intuitively require buying a FitBit tracker to use. Would welcome others' thoughts on this.

agraebe:
I think the FitBit analogy is fair. As for developers, there are many use cases that are not tied to an actual, physical trip. For instance:
- developers can obtain rider profiles, price and time estimates to help decision making for transportation options
- developers can use webhooks to understand intend to ride as well as trip status
- developers can obtain driver profiles (rating, payments, trip history)

For a bit more color on this, all listed transportation APIs are all tied to paid services (train, bus, etc).

mikestreety:
I would agree with @agraebe - as long as the API is free, I see no reason why it shouldn't go in 😄

davemachado:
Good to have a base case for future additions!

PR #274, opened 2017-01-31
Mark Adorable Avatars as HTTPS capable

tathamoddie:
While their main website returns the wrong certificate, the actual API does support HTTPS


mikestreety:
Thanks @tathamoddie

PR #275, opened 2017-02-03
Create Math Section and Add Newton API

aunyks:
In this Pull Request, I intend to add a mathematics section to this API list. This is because the API that I am adding is not accurately described as a Science API (but more of a Math API).


davemachado:
Thanks @aunyks ! :)

PR #276, opened 2017-02-04
Fix two sections not showing links in the readme

rmlewisuk:
[Documents & Productivity](https://github.com/toddmotto/public-apis#documents--productivity) and [Data Validation](https://github.com/toddmotto/public-apis#data-validation) were both missing the `HTTPS` column, so the links weren't rendering in the readme. Docs+Prod also said `OAuth` instead of `Auth` in the header.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ For information on contributing to this project, please see the [contributing gu
 
 ### Documents & Productivity
 
-| API | Description | OAuth |Link |
+| API | Description | Auth | HTTPS |Link |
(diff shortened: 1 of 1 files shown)


davemachado:
Nice catch @rmlewisuk , thanks! 👍

mikestreety:
Ah cheers @rmlewisuk - the reason they were missing the `HTTPS` header was becuase they said oAuth - on this note data validation still says oAuth - will go and fix now

PR #280, opened 2017-02-08
OAuth corrections

jeremiahlee:
OAuth should be capitalized "OAuth" per [RFC 6749](https://tools.ietf.org/html/rfc6749).


mikestreety:
Ha! I accidentally clicked close and it wouldn't let me merge it ;)

PR #287, opened 2017-02-15
Added LectServe to Calendar

marmanold:
LectServe is a Protestant lectionary featuring the Revised Common
Lectionary and the ACNA Lectionary


davemachado:
Thanks!

PR #288, opened 2017-02-16
Add Teleport.org quality of life database API

zackharley:
Excerpt from the Teleport API overview page:

> We're building the largest and most up-to-date quality of life database for Planet Earth using 300+ different data dimensions from 70+ different sources, all updated periodically. Our global sources include the World Bank, World Health Organization, United Nations, Reporters Without Borders, OpenStreetMap, GeoNames, OpenFlights, Heritage Foundation, AngelList, Airbnb, Seed-DB and others. On top of these we augment the sets with things like laws and regulations, real estate prices and recruitment market data from local sources from countries around the world.


davemachado:
Thanks @zackharley !

PR #289, opened 2017-02-16
The Clarifai URL was wrong

stinkyfriend:
Now using the correct URL


davemachado:
Thanks for catching that @stinkyfriend !

PR #291, opened 2017-02-16
Update the help url for upstream remote

stinkyfriend:
I found this GitHub help url more helpful.


davemachado:
Seems like this link does have more info than the other one. Thanks!

PR #296, opened 2017-03-03
Added Unplugg Automated Forecasting API

ruimagalhaes:
Added under Machine Learning


jbrooksuk:
The build is not failing because of your link. Don't worry!

davemachado:
@ruimagalhaes does this API support HTTPS? All the examples in the docs seem to use HTTP.

ruimagalhaes:
@davemachado yes, the API does support https. We are on the process to bring it to the site too but the API is definitely covered. Cheers!

davemachado:
@ruimagalhaes thanks, looks like the endpoint is ` https://api.unplu.gg/`, so looks good to me!

PR #299, opened 2017-03-11
Add geocoding api

zsteeby:
Added Geocode.xyz to the geocoding section.  I set the auth to No, but an `apiKey` is required for unthrottled access.  https://geocode.xyz


davemachado:
Thanks @zsteeby !

PR #300, opened 2017-03-11
Added the Weatherbit API.

cacraig:
Added under Weather header. Thought it would be appropriate to list here!

Thanks.


mikestreety:
Thanks!

PR #302, opened 2017-03-15
Add PayPal

KellyMarchewa:
I just added a link to PayPal's API. I apologize if I made any mistakes - this is my first pull request. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ For information on contributing to this project, please see the [contributing gu
 |---|---|---|---|---|
 | mailgun | Email Service | `apiKey` query string | Yes | [Go!](https://www.mailgun.com/) |
 | markerapi | Trademark Search | No | No | [Go!](http://www.markerapi.com/) |
-| PayPal | Online Payment Processing | `OAuth` | Yes | [Go!] (https://developer.paypal.com/) |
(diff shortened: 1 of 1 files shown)


davemachado:
Hi @KellyMarchewa , thanks for the pull request! After doing a bit of digging into the PayPal API, I was unable to find any type of free services/tier option for the services availible with the API. As this list is a collection of free services, do you know if this API offers any free services?

KellyMarchewa:
Thanks for the response @davemachado. To my understanding, there are no fees associated with using the[ REST API ](https://developer.paypal.com/docs/api/overview/) itself, with the exception of certain features such as accepting credit cards directly. A business account is required, but this is also free. However, there, the standard fees/deductions associated with receiving PayPal payments for business transaction. As such, I certainly understand if this is not a suitable service for the project. 

Thanks regardless, and keep up the great work. This is a very nice resource. :)

davemachado:
Hi @KellyMarchewa , thanks for the clarification. While the API usage may not incur cost, using any sort of service with it will. Because of this, the PayPal API won't be able to find a home here. Thanks for your pull request, and I encourage you to submit new ones in the future! :)

PR #304, opened 2017-03-17
Add Studio Ghibli

janaipakos:
Picked Anime for the section. If there are enough apis to warrant it now or in the future, maybe rename to Animation?


davemachado:
Thanks @janaipakos !

PR #310, opened 2017-03-27
Add ipify

jpalmieri:
Adding [IPify](https://www.ipify.org/) to the list. It's a simple API that just returns your IP address, like so:
```
$ curl 'https://api.ipify.org?format=json'
{"ip":"4.53.140.54"}
```


davemachado:
Thanks @jpalmieri !

PR #311, opened 2017-03-28
Added Victoria's PTV API

oferze:


mikestreety:
Thanks for this @oferze. How does one obtain a developer key? I can't seem to see it! 😄

oferze:
@mikestreety in my last commit (e981e61) I updated the URL to one that actually shows you how to get a key.

mikestreety:
Great, thanks!

PR #324, opened 2017-04-03
Added Utilities section & the first service for it (JSON 2 JSONP)

fevangelou:
Sometimes a service may not provide its data directly in JSONP for use with client-side JS. This high-availability service (already serving more than 7m requests monthly) can assist in converting data output in JSON to JSONP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,6 @@ For information on contributing to this project, please see the [contributing gu
 * [Sports & Fitness](#sports--fitness)
 * [Transportation](#transportation)
 * [University](#university)
-* [Utilities](#utilities)
 * [Vehicle](#vehicle)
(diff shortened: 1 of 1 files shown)


jbrooksuk:
Would this not be better suited under https://github.com/toddmotto/public-apis#development?

fevangelou:
Sure, it was just a recommendation :)

jbrooksuk:
Np!

I think we're better keeping the categories simpler :) I'll merge once this is updated.

fevangelou:
Done.

jbrooksuk:
Oh, I should've updated the merge commit message, but hey ho. Thanks!

fevangelou:
Thank you :)

PR #325, opened 2017-04-03
Added Metadefender Cloud to Malware section

kittrCZ:
I'm adding Metadefender Cloud to the section about Malware.


davemachado:
Hi @kittrCZ , thanks for the pull request!

It appears this service is not a free tier solution, as according to their site:
```
For the free usage you will be able to consume the scan report from 40+ anti-malware engines, but you will not be allowed to use in any way the data provided by our platform. 
```

Can you please confirm this? Thanks.

kittrCZ:
Hi @davemachado, the service is free to use for ["Available only for demo purposes"](https://www.metadefender.com/licensing), if you want to use this API in production systems, you have to obtain Commercial license.

davemachado:
@kittrCZ , thanks for the confirmation! So it looks like this service does not want this data used for anything beyond testing, which doesn't necessarily fit into a "free tier" category. I'd encourage you to open more pull requests in the future though!

PR #333, opened 2017-04-05
Add Zomato API

dakshshah96:
Zomato is a restaurant search and discovery service founded in India. It currently operates in 23 countries, including India, Australia and the United States.


exexzian:
@dakshshah96 do developers really use zomato API, be it company projects and standalone side projects? 
Or I am missing some trending use-case? Will be glad to know.

davemachado:
Thanks!

PR #334, opened 2017-04-05
Fix Goodreads typo

dakshshah96:
Goodreads is currently referred to as Good Reads which is an obvious typo looking at [their website](https://goodreads.com) or practically anywhere else on the web.


davemachado:
Thanks!

PR #340, opened 2017-04-17
Update Schiphol API and improvements

donroyco:
- Schiphol API link was deprecated. This is the new portal.
- Reordered Transportation section alphabetically.


davemachado:
Look good, thanks @donroyco !

PR #342, opened 2017-04-18
Parcel tracking

afabijan:
I've added the parcel tracking category and the parcel service tracking api for Sweden. Maybe this is of interest for someone...


jbrooksuk:
Would this be better under **Data Access**?

afabijan:
Maybe. I don't mind where it is imho. I am planning to add other parcel tracking apis...

mikestreety:
Hmm, I personally wouldn't say it belongs in either... Data access seems to be about Governments, while a parcel tracking category is too specific.

What about a *Tracking* category in general - other trackers could then be added if desired? Be interested to hear what @davemachado thinks. He seems to have a good understanding 😄

mikestreety:
Thanks @davemachado!

@afabijan Are you ok to action? 😄

davemachado:
@mikestreety This is a great example of how having a fuzzy search web interface (issue #203) for this list would be beneficial when a category name might not be enough to find the entry the user is looking for :)

afabijan:
Perfect! Thanks for the input everyone! :) Ok to action.

mikestreety:
@afabijan Just needs to be in alphabetical order now 😉 

:+1:

afabijan:
The build seems to fail, however, I don't think that it's couse of my changes :).

mikestreety:
Thanks @afabijan !

davemachado:
@afabijan sorry just seeing this now - for future reference, the build will fail if one or more linked services are timing out. The build failing will have no impact on the PR's validity as long as the added link(s) aren't breaking the build :)

PR #343, opened 2017-04-19
add the Navitia API

antoine-de:
The [Navitia](https://www.navitia.io/) API is an open API built around transport data.

you can query for journeys, isochrone, schedules, realtime data, ...


mikestreety:
Thanks!

PR #358, opened 2017-05-19
CI Update 

ReadmeCritic:
White listing a few sites that are flagged in error


ReadmeCritic:
hm looks like the errors are gone

PR #359, opened 2017-05-21
Remove OMDB API

arialblack14:
This API has gone [private](https://www.patreon.com/posts/api-is-going-10743518) and now a small fee is required to use it.


mikestreety:
Good spot. Thanks @arialblack14!

PR #360, opened 2017-05-21
Added Numbers API & Meetup API

ApCoder123:
Added the Numbers API [Go](http://numbersapi.com) which gives random facts about numbers & added Meetup API, which gives data about meetups of meetup.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,6 +317,8 @@ Please note a passing build status indicates all listed APIs are available since
 | API | Description | Auth | HTTPS | Link |
 |---|---|---|---|---|
 | Newton | Symbolic and Arithmetic Math Calculator | No | Yes | [Go!](https://newton.now.sh/) |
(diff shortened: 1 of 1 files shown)


davemachado:
Hi @ApCoder123 , thanks for the PR! The Numbers API looks great - thanks for adding.

After doing some digging around, it appears that Meetup [does not offer a free tier for their services](http://blog.meetup.com/fairer-meetup-price-plans/). If so, this service won't be able to find a home here. Could you please confirm? Thanks!

ApCoder123:
@davemachado I added meetup, because the API allows you to get data for free about meetups on meetup.com as well as events, meetups in an area etc. And all you need to get an API key is a **Free** meetup.com account
[Meetup API Docs](https://www.meetup.com/meetup_api/docs/)

Otherwise, just add the numbers api

davemachado:
@ApCoder123  Thanks for the clarification! Moving ahead with both entries - thanks for the PR!

PR #362, opened 2017-05-25
Use CI to generate JSON for all API entries

davemachado:
So, picking up where PR #259 left off, I wanted to find a way to generate beautifully parsable JSON from our Markdown tables. I worked out a Node.js script that will parse through all the various tables and build a JSON object from all the tables. In the deploy script, I create both a minified and unminified version of the JSON. The last little piece of this is setting up Travis to create these JSON files on master builds and push them up to GitHub afterwards. Would love to hear some thoughts on this :)


jbrooksuk:
Could we maybe rename the minified JSON to something else? Better yet, do we need a readable and non-readable version?

davemachado:
The main motivation behind this is to offer a "public api for /public-apis", since a user could simply scrape for the raw JSON file.

davemachado:
There's still talk out there to move/offer this list from a website, so have a non-minified version would make things easier to read/plan if that ever takes off. I'm also fine with also just having one JSON file in the base directory that people can scrape.

davemachado:
@jbrooksuk did you have an idea for the minified json file's name? They're always the classy `data.json` or similar.

davemachado:
I'm going to merge this into a new branch now and open a new PR to merge into master so PR #366 isn't blocked.

PR #364, opened 2017-05-30
changes "free" to "paid" for holidayapi.com since its not free anymore

jitbit:


davemachado:
Hi @jitbit , thanks for the PR. It looks like this service still does offer a free tier, but only for historical data. As such, I've updated the description of the API to follow suit. Thanks!

PR #365, opened 2017-05-31
add gene_info_api

immexerxez:


jbrooksuk:
Looks like the link may need to start with www.

PR #366, opened 2017-06-03
gh-pages branch for a site with search capabilities

kuhnrl30:
I converted the markdown to json and then built a jekyll site on top of it.  Nothing too fancy but accomplishes 2 goals: 

1. Convert the APIs to json format for ease of use. You can get at the data directly with a link like this: https://raw.githubusercontent.com/kuhnrl30/public-apis/gh-pages/_data/apis.json.

2. Add the ability to search across 4 columns- name, description, auth type, and category. 

You can demo the site on my fork of the repo: http://ryankuhn.net/public-apis.  If accepted, we'd have to do a few tweaks to move it to Todd's repo e.g. changing the repo link in right side of the header.


jbrooksuk:
I believe we have a branch with JSON already being generated. It may be worth rebasing your branch on that one.

kuhnrl30:
From what I can tell, that branch is still  a pull request (#362) and hasn't been merged in yet? I can rebase off @davemachado 's fork if that's the better approach.  That way we can continue with that method of generating the json but I'd point the output (or a duplicate of the output) to the _data/ directory in the gh-pages branch.

The json file used in my branch was built by hand with the help of a text editor and regex.  I'm less concerned with how it's generated as long as it makes its way to the _data directory for jekyll to work properly.

davemachado:
Hi @kuhnrl30 - I merged my JSON generation into a new [branch](https://github.com/toddmotto/public-apis/tree/feature/add-json), so you don't need to work off of my own fork.

PR #368, opened 2017-06-03
Bring JSON generation into master branch

davemachado:
Continuing where PR #362 left off.


json/api.json
```diff
@@ -0,0 +1,2178 @@
+{
+    "Animals": [
+        {
+            "API": "IUCN", 
+            "Auth": "`token`", 
+            "Description": "IUCN Red List of Threatened Species", 
+            "HTTPS": "No",
```
yannbertrand: Maybe convert to boolean?

json/api.json
```diff
@@ -0,0 +1,2178 @@
+{
+    "Animals": [
+        {
+            "API": "IUCN", 
+            "Auth": "`token`",
```
yannbertrand: What about removing "`" chars and set its value to null if not auth required?

build/md2json.js
```diff
@@ -0,0 +1,116 @@
...
+        } else {
+            console.error("markdown expect table content")
+            break;
+        }
+
+        line = read_line()
+
+        if (line) {
+
+        } else {
```
yannbertrand: if (! line) {
      console.error('markdown expect table spliter')
    }

yannbertrand: Not tested but should work

davemachado:
@yannbertrand thanks for the review! I updated `md2json` accordingly, so it now sets auth to null if no auth is needed and HTTPS support is set to boolean values. I think that silly `if (line) { }` statement was supposed to be a placeholder for something once upon a time, but it's obviously not needed anymore.

I also updated the stock json files to reflect the new format. Let me know what y'all think!

mikestreety:
I do agree with @kuhnrl30 - I personally think categories should be an array on each item. When rendered out, we can display as they currently are (then that means a link can appear in many categories, but only have one source of truth).

If you wanted to go down the "programming" route, each category could have an ID, to stop typos and very similar categories appearing.

e.g.

```
{
	"Categories": [
		1: "Animals",
		2: "Anime"
	],
	"APIs": [
		{
			"API": "AniList", 
			"Auth": "`OAuth`", 
			"Description": "AniList Anime", 
			"HTTPS": "No", 
			"Link": "http://anilist-api.readthedocs.io/en/latest/",
			"Categories": [1, 2]
		}
	]
}
```

davemachado:
I should clarify my opinion - for a website with searching capability, I think categories make complete sense, since the information can easily be filtered for viewing without being overwhelming. However, the list of entries currently stands as one big README, which means we'd need duplicate listings in the one file. That would also mean that any updates to an entry (i.e. new auth requirements) would need to hit all instances of the entry in the file. I agree with @mikestreety 's point of having one source of truth, but with the current layout of the README, I don't see how that's going to be possible for sane maintenance.

[twofactorauth](https://github.com/2factorauth/twofactorauth/tree/master/_data) is a nice project that many people base these types of "list websites" off of, and they use YAML files for categories to render each independently of each other, only coming together in the final rendering. Without trying to overcomplicate our project here, I think we would need to adopt a variation of the separate files system in order to reduce overall maintenance of entries without duplication of effort.

More food for thought :)

kuhnrl30:
Do we even need the README to be a source of truth?  Can we convert to JSON once and then use that as the source going forward?  The README could link to the site for easy reading.  

You can use json or yaml in jekyll site. We've already started with json so I'd suggest we stay with it.

davemachado:
This project has quite a large footprint in terms of daily GitHub traffic, so we need to proceed with caution and respect for the visitors here - baby steps are key for smooth transitions. The README is the key component to the overall use of the project right now. After incorporating the JSON into the repo, I think a discussion of possibly updating the source of truth to be the JSON generating the Markdown tables can be warranted, but the first step is to bring this new structure into the fold.

davemachado:
I am going to start taking this in a slightly different direction - will have a new PR posted soon.

davemachado:
Not sure I closed this...but anyways, after playing with this a bit, It's becoming clear how much of a nightmare for maintenance using category IDs would be. For example, if someone wants to create a new category, they'll have to find the last known ID number used, which is going to be out of order (i.e. not at the end of the list) depending on if the categories are kept in alphabetical order. Currently, users sometimes have a hard time putting their new entries in alphabetical order, and this change in workflow would only exacerbate it.

I would like to move ahead with the originally proposed JSON that reflects the current format of the README, which is the source of truth for the entire project. From there, I think separate conversations can be started regarding any thoughts for migrating the source of truth from the README to the JSON, website, etc, which could then include motivation to change the format of the JSON, and therefore the README.

However, the goal of the feature outlined in this PR is to provide an easy-to-code-with format of the project. Since the project is based on the category-sorted markdown tables in the README, I'm standing by my original proposal as both "good enough for now" and to inspire motivation to start those separate conversations mentioned above.

As always, thoughts are welcome.

PR #369, opened 2017-06-04
Fix Kitsu API docs link

wopian:
`kitsu17` was a temporary apiary page - its [now](https://github.com/hummingbird-me/api-docs) just `kitsu`.

Also fixed the `HTTPS` column as it has always been HTTPS-only


wopian:
Travis failure is due to [Steam's API](https://developer.valvesoftware.com/wiki/Steam_Web_API) currently 500ing

davemachado:
Thanks!

PR #370, opened 2017-06-09
Wunderground edit

reifnotreef:
Wunderground requires a key.
Developers can get a free one, if/once registered, that has daily/minute limits.


davemachado:
Thanks @reifnotreef !

PR #372, opened 2017-06-12
Added more APIs

RonildoSouza:


MrPowerGamerBR:
> The Cartola FC API serves to check the partial points of your time.

Shouldn't it be "team" instead of "time"?

PR #378, opened 2017-06-26
Fix --allow syntax

DuckThom:
HTTP Codes in the `--allow` option are comma separated.
This should fix the following error in the builds:

```
File open error: No such file or directory @ rb_sysopen - 302

Summary

README.md: ✓

      302: No such file or directory @ rb_sysopen - 302

The command "awesome_bot README.md --allow-ssl --allow 403 302" exited with 1.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: ruby
 rvm: 2.2
 before_script: gem install awesome_bot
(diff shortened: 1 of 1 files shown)


davemachado:
Nice catch, the comma must have slipped by me while updating the build config.

PR #379, opened 2017-06-27
Added more APIs for ML

hardikpthv:


davemachado:
Hi @hardikpthv , thanks for the PR! Unfortunately, both Google Cloud and Amazon Web Services only offer temporary free tier services, so they won't be able to find a home on this list. Otherwise, we'd have 50+ AWS entries for all their different services :)

PR #385, opened 2017-07-10
Add Format Validation to CI

davemachado:
So, right now the CI for this project tests every link in the README to make sure that it's living. This is great for removing dead links over time, but can cause a real headache for users who open a pull request only to see their changes "break" the build, when 99.9% of the time it is someone else's link that is causing the failure.

Therefore, in the spirit of alleviating this headache, I have written a simple format validator that will run on pull requests. This validation checks for valid auth values, markdown link syntax, and a few other goodies.

I plan to add more checks to this soon, but my plan is to merge in the CI changes quickly to stop having pull requests fail for no good reason - and to give maintainers a break from having to point out simple fixes :)


davemachado:
To clarify, the link validation will still run on the master branch, where it really is intended to be run. Just not on every single pull request - I'd like to think that someone submitting a new entry would not be adding a dead link....

build/main.sh
```diff
@@ -0,0 +1,16 @@
...
+./validate.rb ../README.md
+
+if ["$?" == 0]; then
+    echo "format validation PASSED"
+else
+    echo "format validation FAILED"
+fi
+
+if [ "$TRAVIS_BRANCH" == "master" ]
+then
```
jbrooksuk: Should this not be on the line above, like the rest of the conditions?

PR #387, opened 2017-07-21
Add open government data for New Zealand

aquibm:
Add open government data for New Zealand 🇳🇿 
[https://www.data.govt.nz](https://www.data.govt.nz)


davemachado:
Looks good, thanks!

PR #388, opened 2017-07-21
Validate PR updates exclusively

davemachado:
This amends the way CI approves a PR. Currently, a PR CI build will fail if any link is invalid when the PR is opened, resulting in many PR failures being false negatives. With the update, CI will only check links modified by the request, not all. This should result in accurate CI failures. If no one has objections, I will be merging this in to master-branch over the weekend.


jbrooksuk:
Very clever!

PR #390, opened 2017-07-26
Added Blackbaze B2 api

pyprism:


jbrooksuk:
Is Backblaze free? I don't think this fits into this repo.

apearson:
First 10GB are free on the service.

jbrooksuk:
I'm still not sure it fits into this service. My understanding is that we're only sharing usable services? @davemachado?

davemachado:
Hey guys, thanks for bringing this up. Once upon a time, I looked at using BackBlaze for a small backup solution. The problem with the free 10GB tier is that you still need to give 'em your credit card just in case you happen to go over the 10GB limit while backing up your terabytes of homelab data ;)

In my opinion, that need to give out financial information goes against the spirit of this project, so I'd recommend holding off on adding BackBlaze.

pyprism:
I recently tried their B2 free plan without giving any financial info .

jbrooksuk:
At the very least it required a phone number.

mikestreety:
I think we might have to say no to this i'm afraid. I just tried to sign up and got bombarded with a survey - plus I can't find any info on thier free tier on thier site.

Thank you anyway @pyprism - look forward to seeing your next PR :)

PR #405, opened 2017-07-26
Move SWAPI to Video section

ianwelsh:


davemachado:
Hi @ianwelsh , thanks for the PR! Can you explain why you think this service belongs in `Video`?

ianwelsh:
While there are Star Wars games and comics, the franchise is centered around the film series. 
Since the `Video` section has TV and film franchise inspired API's (An API of Ice And Fire, Ron Swanson Quotes) then the Star Wars API belongs there as well.

davemachado:
Great, thanks!

PR #421, opened 2017-07-27
Build changes

jbrooksuk:

PR #426, opened 2017-07-28
Add new format validation check and update logic

davemachado:
Seeing as though @jbrooksuk beat me to fixing some ugly parts of my markdown-to-json script in #421 , I'm offering up a new sacrifice (and maybe I wanted to play with @mikestreety 's new labels...)

This PR covers the following:
- Add a check to make sure `Description` is not punctuated. Most of these are just blurbs rather  than elegantly crafted English prose.
- Add a check to make sure each segment is padded by exactly 1 space
- Stop using a stupid variable (`fail_flag`) that is set over and over again and instead store errors in an array to be printed at the end. The length of the array can then be referenced to indicate the number of errors (hopefully 0!)
- Now that we have access to the number of errors, why not set the exit code to that number? In the case of `n` number of errors, the bash script that calls the validator will have access to the number of errors if it ever wants to reference that in the future.
- minor structure changes to make maintaining and extending this easier over time

Let me know what you guys think and if you have any other ideas, I'm all ears!


davemachado:
Note - I did not update the README on purpose so you can look at the CI build for these and see what the new output looks like. Once this gets the green light, I will update the README to conform to the checks.

jbrooksuk:
This is nice!

davemachado:
@jbrooksuk is that a "wow that's great you should totally go ahead and merge that in!" comment? ;)

PR #432, opened 2017-08-01
News

edsu:
I added an entry for a JSON API for historic newspapers from the Library of Congress.

Thanks for a great resource that is very handy for teaching folks about APIs.


jbrooksuk:
Thanks! :)

PR #436, opened 2017-08-02
Updated Musikki entry

the-lay:
Musikki requires AppKey and AppId that can be obtained in the Musikki dashboard


the-lay:
Source: https://music-api.musikki.com/overview#quickstart

mikestreety:
Thanks!

PR #418, opened 2017-07-27
Updated APIs list with ADS-B flight data API

timtheguy:
I have added the API provided by the ADS-B Exchange, giving access to both real-time and historical information for flying aircraft. The API is [available at the ADS-B exchange website](https://www.adsbexchange.com/data/), free of charge for non-commercial use.

I did not feel that any available categories aligned with the services offered through this API. 

For the "Vehicles" category, though yes, airplanes are vehicles, this API is not offering a service to learn more about specific vehicles or get more information regarding a certain aircraft model.

For the "Transportation" category, the existing services listed all relate to transportation readily available to passengers, such as those related to personal travel. This API details information for all kinds of aircraft, including military and private training aircraft.

PR #438, opened 2017-08-02
Adds section 'Travel' and APIs.is (Collection of Icelandic APIs)

zebroc:
Adds section 'Travel' and APIs.is (Collection of Icelandic APIs)

I think it makes sense to add a 'Travel' section, there surely are other APIs out there that fit in it

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ Please note a passing build status indicates all listed APIs are available since
 * [Sports & Fitness](#sports--fitness)
 * [Tracking](#tracking)
 * [Transportation](#transportation)
+* [Travel](#travel)
 * [University](#university)
 * [Vehicle](#vehicle)
 * [Video](#video)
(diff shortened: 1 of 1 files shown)


jbrooksuk:
We've put a few Travel related API's under Transportation. I wonder if we should rename it to Travel/Transportation or do as you've done here and separate them?

Ping @davemachado @mikestreety

mikestreety:
Travel is transportation?! 😉 I'd be in favour of either:

- Separating but calling it Public Transport
- Keeping them together with a slash

davemachado:
I could have sworn I added more to address this idea to my response in #418 .... guess not!

IMO, I would recommend `Travel & Transportation` for now - if we start receiving an influx of PRs for that section and there is enough meat for both `Travel` AND `Transportation`, then and only then would splitting make sense (this idea would still hold for any section, such as if we ended up with enough `Food` entries AND `Drink` entries to warrant splitting `Food & Drink`). Like I mentioned in the linked PR, I would hate to see a whole bunch of mini-sections start popping up.

mikestreety:
Yup, I'm with @davemachado.

@zebroc, would you mind popping this addition into Transportation?

zebroc:
Done.

jbrooksuk:
Perfect! Thanks :)

PR #440, opened 2017-08-03
Build Bot comments on failed builds

davemachado:
So before I get working on the site plan, I wanted to throw this idea out. I wrote a bot that will comment on a pull request if the build fails, giving the user a notification/email that their submission failed, instead of having to wait for a poor maintainer to come along and review. You can see an example on my fork [here](https://github.com/davemachado/public-apis/pull/4).


davemachado:
The Build Bot program can be referenced [here](https://pypi.python.org/pypi/build-bot)

jbrooksuk:
Could we setup a user for Comments?

davemachado:
Well I did a bit more digging and I'm thinking to actually scrap all this work in favor of just enabling emails to be sent by Travis to the contributor in the event of a failure.

If that doesn't work out, yes we could easily set up a bot user to manage comments.

jbrooksuk:
Let’s try that option first :)

davemachado:
Will enable emails as a first line of defense 🤺

PR #441, opened 2017-08-03
Update JSON to set Section as attribute

davemachado:
Using the JSON format to generate a web table will be a lot easier if the JSON is not structured in a section-based hierarchy. Instead, the JSON should simply be an array of entries with the corresponding Section set as an attribute. I also threw a total entry count into the mix for good measure.

The way to accomplish this without mucking with my existing md2json generator is to simply make the README into one big Markdown table (`single_table.md`), which is then passed through to the generator. This is done in the updated build script.

The PR has both the JSON and condensed files included to better show the changes.


jbrooksuk:
Should we do a cleanup and remove some of these created files afterwards?

davemachado:
I had thought this over while working on the condensation script and since it is only one file that is generated, I figured the convenience of being able to check that file in the build if something needs to be debugged outweighs the formality of cleaning up after oneself.

PR #444, opened 2017-08-04
Updated Mixcloud entry

the-lay:
Mixcloud requires OAuth authorization


the-lay:
Source: https://www.mixcloud.com/developers/#authorization

mikestreety:
Good catch - thanks

PR #447, opened 2017-08-06
Update JSONbin.io - Now accepts API Key for creating private records.

i-break-codes:


jbrooksuk:
I wonder if we need to denote optional api keys?

i-break-codes:
@jbrooksuk Would love to do that, I didn't see any option to do so :)

jbrooksuk:
Maybe a question for @davemachado.

davemachado:
Thanks for bringing this up guys. I've run into this a few times, and unfortunately it's not a clear-cut problem with a clear-cut answer. With projects like this one, where code is not the main motivator for PRs and issues, keeping a consistent precedence is key. For example, we've recently seen a lot of people wanting to contribute new entries in new sections since the existing sections don't describe the entry 100%. If we had let any of those PRs merge with single-entry sections, we would have set a precedence that having one-off sections is acceptable.

Now back to this issue - I'm suggesting to leave it how it is right now, using the vanilla `apiKey` label. I've made an offline note that this entry is has an optional Auth. If we start seeing more and more entries have optional Auth, then we could look to add a new label option. The other school of thought for this is that the Auth label should show the "worst-case" option, meaning that in the most advanced use case of a service, I'd have to use the noted Auth. Anything less, such as using a part of the service that requires no Auth, is free cake at that point. 🍰 

TL;DR - 1) We don't want to set a precedence of using a new label until multiple entries need it. 2) Leave it in the "worst-case" scenario of `apiKey` and see how it goes. 3) Dave rambled on here so if this comes up in the future he can reference this.

mikestreety:
I would agree that having worst case scenario is best!

mikestreety:
I pressed the wrong button. Sorry 😉

PR #462, opened 2017-08-17
Alpha website candidate

davemachado:
I've uploaded my alpha site candidate to an S3 bucket, viewable [here](https://s3.amazonaws.com/public-apis-alpha/index.html). It utilizes Vue over JQuery and Bootstrap, minimizing libraries and crazy amounts of minified file fun. I'm more than happy to listen to suggestions on this plan, but I am going to keep pressing forward in this direction. There is still plenty of work to be done here, such as loading the JSON file locally to speed up the table rendering.


yannbertrand:
Good work :smile:

First simple suggestion: fix the column width to avoid column resize on search.

Thank you for pushing this idea forward :smile:

jbrooksuk:
> First simple suggestion: fix the column width to avoid column resize on search.

Definitely!

Few suggestions:

- Store the JSON in local storage
- Add a manifest for offline usage (useful if you're on a train looking for something), similar to DevDocs.io

I'll review the code later on :)

src/index.html
```diff
@@ -0,0 +1,55 @@
+<!doctype html>
+
+<html lang="en">
+	<head>
+	 	<meta charset="utf-8">
+ 		<meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <link rel="stylesheet" type="text/css" href="styles.css"/>
+        <link rel="shortcut icon" href="favicon.ico" type="image/x-icon"/>
+		<script src="https://unpkg.com/vue"></script>
```
jbrooksuk: Let's use the minified version of Vue https://unpkg.com/vue@2.4.2/dist/vue.min.js

src/scripts.js
```diff
@@ -0,0 +1,72 @@
+ loadJSON(function(response) {
```
jbrooksuk: We shouldn't need to wait to load the JSON before the page kicks in, this will slow down the initial rendering. I'd say we opt for promises too.

src/scripts.js
```diff
@@ -0,0 +1,72 @@
...
+          if (xobj.readyState == 4 && xobj.status == "200") {
+            // Required use of an anonymous callback as .open will NOT return a value
+            // but simply returns undefined in asynchronous mode
+            callback(xobj.responseText);
+          }
+    };
+    xobj.send(null);  
+ }
+
+function filterRows() {
```
jbrooksuk: I have a feeling this can be done with Vue itself. Leave it with me, I'll pull down the branch and have a play later.

src/styles.css
```diff
@@ -0,0 +1,48 @@
+body {
+    border-collapse: collapse;
+    font-family: "Trebuchet MS", Arial, Helvetica, sans-serif;
```
jbrooksuk: Can we use a nicer font-stack, maybe something from Google Fonts?

src/scripts.js
mikestreety: I've not looked at the code too much, but could the filtering be done like this?

https://codepen.io/mikestreety/pen/XgYZEy

src/scripts.js
mikestreety: YOu can use the `created()` function on the Vue instance itself to load the data, so everything thing else can initialise while you wait.

Like this 😄  https://codepen.io/mikestreety/pen/jLYGbW

src/index.html
mikestreety: Also the JS should be in the footer ;)

mikestreety: Also the JS should be in the footer ;)

mikestreety: Also the JS should be in the footer ;)

src/scripts.js
mikestreety: YOu can use the created() function on the Vue instance itself to load the data, so everything thing else can initialise while you wait.

Like this 😄 https://codepen.io/mikestreety/pen/jLYGbW

src/scripts.js
mikestreety: I've not looked at the code too much, but could the filtering be done like this?

https://codepen.io/mikestreety/pen/XgYZEy

mikestreety:
Sorry if everyone got loads of notifications - I cocked up the review ;) Sorted now!

src/scripts.js
jbrooksuk: Exactly this!

src/scripts.js
jbrooksuk: Yes!

src/index.html
jbrooksuk: Or at least `async` ;)

davemachado:
Thanks for the great feedback everyone! I've made the following improvements:

- fixed column width (may want to tweak a bit so longer string columns get more space)
- added AppCache manifest
- Switched to Roboto font from Google Fonts
- Use Vue to filter search results and fetch JSON (still want to look into a better JSON delivery method)

I will update S3 as well.

src/index.html
```diff
@@ -22,10 +23,10 @@ <h3 class="center-text">
 		</h3>
 	</header>
 	<body>
-		<div class=center-text>
-			<input type="text" id="searchbox" onkeyup="filterRows()" placeholder="search for a service">
-		</div>
 		<div id="app">
+			<div class=center-text>
```
jbrooksuk: Missing quotes.

src/index.html
```diff
@@ -38,18 +39,20 @@ <h3 class="center-text">
...
-							<td class="lol">{{ item.Category }}</td>
+							<td>{{ item.Category }}</td>
 						</tr> 
 					</tbody>
 				</table>
 			</div>
 		</div>
-	 	<script src="scripts.js"></script>
 	</body>
+	<footer>
```
jbrooksuk: We don't need these tags.

davemachado:
Side note - I've never used Vue.js before, but I'm liking the functionality so far!

davemachado:
Merging this in to start second iteration.

jbrooksuk:
:)

PR #463, opened 2017-08-17
Add Mercury API

abialbon:
Mercury API takes an URL and return the parsed HTML in a JSON format. Eg. headline, author, body text, relevant images and more — free from any clutter. It’s free. Uses apiKey and requires opening an account to get the key.


mikestreety:
Looks good! Thanks :)

PR #477, opened 2017-09-03
added freecodecamp weather api

hellotunmbi:
I added a new Weather API  from freecodecamp.


davemachado:
Hi @hellotunmbi , thanks for the PR! Unfortunately, this endpoint is really just a passthrough, which is not meant to be used outside of the freecodecamp course scope. Rather, it appears this service exists to allow students to practice using APIs in their apps. Because this is just a mask rather than a service itself, it won't be able to find a home in this project. 

Feel free to submit more entries in the future :)

PR #496, opened 2017-10-03
good reads seems to require api keys now

juliankrispel:
Proof: 
image


jbrooksuk:
Thanks for the update!

PR #528, opened 2017-10-26
JSON is auto-generated, don't touch

jbrooksuk:
This may help.

PR #530, opened 2017-10-26
Add check to see if /json was changed

davemachado:
Following up to PR #528 , adding a check to the build process to see if any files were touched and fail if so.

Adding a /json here to test the build's new code and will revert before merging.


mikestreety:
I don't understand travis...so yeah? 🤷

davemachado:
I'll take it 🍰

PR #548, opened 2017-11-15
Create an API for an API-based project

davemachado:
So, this project is all about collecting public API services from all corners of the internet, yet does not offer its own API... no longer!

During my work on create a website, I figured having a nice little API service would be helpful as well. Currently, the only endpoint allows our monstrous collection to be filtered down by title, description, auth type, etc. 

Figured this would be a good starting point, and I am more than happy to add more features if folks have any ideas. 

Feel free to give it a try [here](http://api.davemachado.com/api?category=development&https=true&auth=apiKey&description=screenshots).

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ API | Description | Auth | HTTPS | Link |
 | Church Calendar | Catholic liturgical calendar | No | No | [Go!](http://calapi.inadiutorium.cz/) |
 | Czech Namedays Calendar | Lookup for a name and returns nameday date | No | No | [Go!](http://svatky.adresa.info/) |
 | Google Calendar | Display, create and modify Google calendar events | `OAuth` | Yes | [Go!](https://developers.google.com/google-apps/calendar/) |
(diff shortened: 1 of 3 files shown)


henhouse:
Hi, yay! Makes me happy to see this written in Go. May I ask why you did not use the standard library for the HTTP handling? The standard library is very capable of almost all standard uses at good performance. I would only consider third party libraries when the needs were great and additional functionality was required.

There are a few errors not handled as well, though unlikely to occur, may need to be caught. I'll make a fork to show an example!

davemachado:
Hi @henhouse , always happy to meet another gopher! 😄 

I wrote this initial version as a development service I could use while working on a web interface for this project. As such, I definitely cut some corners while writing it (such as error catching) and used Gin to simply get something working fast.

Thanks for the PR on my fork - I will take a look as soon as I get the chance!

davemachado:
I'm going to spend some more time working on this properly and then revisit integration. I shall return...

PR #559, opened 2017-11-30
Removed dead Coinbin link

mildrenben:


davemachado:
Hi @mildrenben , thanks for the PR! I know a lot of cryptocurrency servers have been taking a huge server hit recently thanks to Bitcoin's 🚀 ride, so let's give Coinbin some time to recover before cutting them off. I will recheck this in a couple days.

mildrenben:
Good call. I didn't even think of the possibility the servers were just getting thrashed.

Side note - would be great to have a tool that ran to check server responses from all these APIs to weed out dead services. Might get a little annoying with the authorised ones, but for the no auth ones it could be worthwhile, maybe something that runs on a pre-push hook?

davemachado:
something like this? 😉 
https://github.com/toddmotto/public-apis/blob/master/build/validate_links.rb

mildrenben:
😅  must have missed that!

davemachado:
After some Google-fu, it appears that Coinbin has gone under (for now...). Moving ahead with removing.

Issue #303, opened 2017-03-16
Is it possible to add a column to show if APIs are CORS enabled?


tublitzed:
If this is something you wouldn't mind me doing, let me know and I'll open a PR for it. I figured I'd checkin here first to make sure this is something you'd be interested in before I do the work:)

I absolutely love this list and refer to it all the time, by the way. Awesome work putting this together.

mgrigajtis:
+1 on this!

mikestreety:
I'm not so familiar with CORS - mind explaining in laymans terms?

mgrigajtis:
Able to make cross domain AJAX calls with client side JavaScript.

maddenamy:
+1 on this!

mikestreety:
Is there anyway we can automate/assume a yes/no without manually checking them all?

Stroemgren:
You sure could, but you would need the API urls and not just the website urls as the list have at the moment.

In the case of having these urls, you could loop over them, making a request and checking the Access-Control-Allow-Origin header.

jbrooksuk:
It's not fun, but we could add the column then retroactively update when we get time?

alemesa1991:
That would be very helpful +1

afabijan:
+1

Tinee:
@tublitzed, did you pick it up? Otherwise, I could take a stab at it.

mikestreety:
@Tinee I would say go go go.

Would suggest you make the default TBC or 🤷‍♀️ as a placeholder?

davemachado:
@tublitzed @Tinee did one of you start working on this? Would hate to see you both duplicate work.

Tinee:
Go for with @tublitzed :)

GoToLoop:
I wonder whether a JSONP option should be included as an option as well.  :confused:

wordythebyrd:
Came here to ask for this.

mikestreety:
Anyone is more than welcome to pick this up :)

torzuoliH:
Would be nice !

davemachado:
Echoing @mikestreety 's comment that anyone is _encouraged_ to start working on this 🥇

mikestreety:
Ok, looks like I need to tackle this. For now, I'll add an empty space or a something.

For people who don't know what CORS is, we should add a "how to check" or similar to the contributing guide.

So, how do you check if it's not obvious in the documentation of the API?

davemachado:
[...]

PR #589, opened 2018-01-14
Add CORS column

davemachado:
Adding a column to all tables indicating if CORS is supported for a service to fix issue #303 .

I have all entries set to `Unknown` to start. I can see an argument for starting with everything set to `No` as well, as to remove a scapegoat choice (i.e. everyone opening pull requests to add new services leaves it as `Unknown` because they don't want to dig into documentation).

Open to hearing opinions either way 😄


jbrooksuk:
LGTM!

mikestreety:
Yes!

davemachado:
Thanks guys!

PR #632, opened 2018-03-19
Adding Instagram Proxy API

whizzzkid:
This is something I wrote almost an year ago. It wraps around Instagram's public API and provides access to public posts programmatically in a CORS compliant fashion. This works really well for building instagram image galleries as it does not require registering an app with instagram. It also provides a one-click deploy with heroku which let's the users host a private instance of the API.

As of today no other solution provides such access to instagram.


davemachado:
Thanks!

PR #647, opened 2018-04-10
Darksky no CORS

technicolor1:
Darksky disabled CORS for security https://darksky.net/dev/docs/faq

>*Why do I get the error No 'Access-Control-Allow-Origin' header is present on the requested resource when I try to call the API?*
We take security very seriously at Dark Sky. As a security precaution we have disabled cross-origin resource sharing (CORS) on our servers.


davemachado:
Thanks!

PR #690, opened 2018-06-08
Fixed Twitter information

andypiper:
Fixed link to Twitter development site and clarified that the API does not support CORS


davemachado:
Awesome, thanks!

PR #700, opened 2018-06-14
Add Bin Codes

barsovichok:
API for checking card number data


davemachado:
Hi @barsovichok , thanks for the PR. When I went to check the link, I was greeted with multiple ads/spam/pop-ups. Given the ads and sensitive nature of the data being collected here, I do not feel confident that this service is reputable.

PR #705, opened 2018-06-27
Update Clash of Clans API

paramt:
Update `Games > Clash of Clans > Auth` from *No* to *`apiKey`* because it is outdated.


davemachado:
Thanks!

PR #728, opened 2018-07-26
Make API names links, remove Go! column

jezmck:
as per issue #727

PR #741, opened 2018-08-12
Update Format Validator to support Title links

davemachado:
Supporting Commit to PR #728. 

This moves the validation logic onto the Title segment rather than the (now-defunct) Link segment.


mikestreety:
Looks good 👍

davemachado:
Covered in e2b66cea263d77107000855ce66fb3554922dd02

PR #765, opened 2018-10-04
Czech personal identification numbers validation

xnekv03:
Adding new API for validation Czech personal identification numbers (rodna cisla) an equivalent of Social security number in US. Applies to Czech and Slovak numbers only.


jbrooksuk:
I’d add in the description that this applies to Czech numbers only.

PR #773, opened 2018-10-09
Updated Clarity Logo and removed Certly

gitoct:
Updated two APIs:

- Added 'apiKey' Auth to Clarity Logo
- Removed 'Certly' because the link no longer works


jbrooksuk:
Can you please do this in two commits.

PR #784, opened 2018-10-15
Remove use of Oxford commas

jbrooksuk:
@davemachado we don't need to use Oxford commas in any of these descriptions 👍


davemachado:
works for me, thanks! 😄

damenleeturks:
Boo! What do you have against added clarity in run-in lists?

jbrooksuk:
Consistency 😂

Issue #792, opened 2018-10-19
Transport for Boston, US - Dead Link


RDumais:
Currently, it points to [https://mbta.com/developers/Home/Documents](https://mbta.com/developers/Home/Documents) 
The correct URL is [https://mbta.com/developers/v3-api](https://mbta.com/developers/v3-api)

davemachado:
Fixed in 57ec644665f74e023cfea5d4945069059357dbee

PR #793, opened 2018-10-19
Updated Transport for Boston, US

RDumais:
Updated Transport for Boston, US with a working URL.
This will resolve issue #792


davemachado:
Thanks!

PR #822, opened 2018-11-19
Add in GeoJS to public APIs

jloh:
Add in GeoJS to Geocoding APIs. This API (created by me) is free to use with open CORS + full HTTPs support. It also has ChatOps integration.

Thanks for creating this awesome list and sing out if there are any issues with my PR!


davemachado:
Thanks!

PR #842, opened 2018-12-20
Added Football Prediction API

boggio-analytics:
Added `Football Prediction API` under the `Sports & Fitness` section.


jbrooksuk:
Thanks!

PR #862, opened 2019-02-04
Postcodes.io does indeed permit cross origin requests from `"*"`

cblanc:
Updates record for postcodes.io to show it does accept CORS requests


davemachado:
Thanks!

PR #886, opened 2019-02-10
Remove street layer because it's going to get discontinued

dooven:
Street Layer API will get discontinued soon: 

remove

So might as well remove it.


jbrooksuk:
Thanks!

Issue #915, opened 2019-03-16
What does ultimatecourses.com do to sponsor public-apis?


DarrienG:
I really like this project and it's great we have all of these APIs in one place, but I'm wondering how Ultimate Courses sponsors this repo.

The APIs are all free, and Digital Ocean provides hosting space, so I'm confused about how Ultimate Courses fits in. It even gets a full banner ad, but Digital Ocean is just a small link?

Does Ultimate Courses pay the contributors? @toddmotto 

I feel like the repo should just be about APIs and if this is just an ad, it pollutes the repo.

DarrienG:
????

DarrienG:
What does this contribute to the repo?

PR #916, opened 2019-03-16
Remove Ultimate Courses ad

DarrienG:
Remove Ultimate Courses ad. We can keep it if it does something for the project or contributors, but it just seems to be taking up space. 

Relevant issue:  #915

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,6 @@
 
 A collective list of free APIs for use in software and web development.
 
-Sponsor:
-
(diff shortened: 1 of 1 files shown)


toddmotto:
I'd like to keep the banner, thanks.

DarrienG:
I'm interested in having a discussion about this. Could we move this to the issue?

https://github.com/toddmotto/public-apis/issues/915

DarrienG:
I'm glad we could have a good discussion about this.

PR #1158, opened 2020-01-29
Fix the link to Transport for Berlin, Germany API

ck3g:
Replaces broken link with its new version.

It seems like there is a new version of the API and docs. The old link is outdated.


yannbertrand:
Thank you!

PR #1214, opened 2020-04-14
Add PIPL API

narbehaj:
PIPL API is a free service that generates random fake person data in JSON. Excellent for data analysis tests and data gatherings.


yannbertrand:
Thanks

PR #1244, opened 2020-04-24
Removed deprecated API Search.gov Jobs

armpatch:
"Search.gov Jobs" API has been deprecated, so I removed it from the list


SuperSandro2000:
What about replacing it with https://developer.usajobs.gov/ ?

armpatch:
That api is already on the list.

matheusfelipeog:
Thanks for contribution @armpatch, but this API has already been removed. 😅

PR #1332, opened 2020-08-12
Russian State Duma Open Data

nikolai-neustroev:
Public API of the lower house of the Russian Parliament


marekdano:
Thanks for the contribution, @nikolai-neustroev!

PR #1345, opened 2020-08-20
Added Etsy API

ellgreen:
Adding Etsy API. I did see issue #519 from a while back that was closed due to Etsy not having a free tier.

I'm not sure if it has changed since then but you definitely make a limited number of requests against the API for free now.


PawelBorkar:
Yeah!
Thanks for the contribution @ellgreen  :)

PR #1365, opened 2020-09-08
Added flight data

aditi-n21:


matheusfelipeog:
Thanks for your contribution @aditi-n21, But I didn't find the documentation for this API. 😅

Issue #1370, opened 2020-09-20
Let's Validate API is defunct and should be removed


t0xic0der:
The "Let's Validate" API has been last worked upon back in 2018. Their website and the API both give a 404 error now which means that the project is probably not maintained anymore.

Rolv-Apneseth:
Made a pull request removing the Let's Validate API entry from the readme. #1378

Rolv-Apneseth:
> Made a pull request removing the Let's Validate API entry from the readme. #1378

First time making a proper pull request, please let me know if I've done things correctly. Thank you

yannbertrand:
It was great, thanks @Rolv-Apneseth and @t0xic0der

t0xic0der:
Glad to help. :smiley:

PR #1378, opened 2020-09-28
Removed the entry for Let's Validate API, as suggested by Issue #1370

Rolv-Apneseth:
Simply removed the entry for Let's Validate API from README.md, as it does appear the api is defunct.

The last update to their github page https://github.com/letsvalidate/api appears to have been 3 years ago and this issue https://github.com/letsvalidate/api/issues/14 open on their github shows the service is down and the api is not being maintained.


yannbertrand:
Thank you @Rolv-Apneseth

PR #1415, opened 2020-10-13
Added saidit.net

FormerlyChucks:
saidit is an open source reddit clone. the api is just like reddit's, which is already on this list.


yannbertrand:
Thanks

PR #1427, opened 2020-10-19
fix: :wrench: adjust user-agent to real value

pyaillet:
change `user-agent` to real value (python-httplib2 and version from `requirements.txt`)


yannbertrand:
Thanks

PR #1428, opened 2020-10-19
chore: :wrench: update user-agent

pyaillet:
update httplib2 version in user-agent after dependabot upgrade

### DIFF
--- a/build/validate_links.py
+++ b/build/validate_links.py
@@ -24,7 +24,7 @@ def validate_links(links):
     for link in links:
         h = httplib2.Http(disable_ssl_certificate_validation=True, timeout=25)
         try:
-            resp = h.request(link, headers={'user-agent': 'python-httplib2/0.9.2'})
+            resp = h.request(link, headers={'user-agent': 'python-httplib2/0.18.0'})
             code = int(resp[0]['status'])
             # check if status code is a client or server error
             if code >= 404:


yannbertrand:
Oh yea thanks

SitramSoft:
I tried to use the real user agent from the lib but I still got 404 errors on some websites. From what I could figure out, some websites require a knows user agent, else they return a 404 error. I checked the logs and the 404 errors are back.

Perhaps PR #1428 and#1427 should be reverted.

yannbertrand:
Can you confirm we're talking about https://www.petfinder.com/developers/v2/docs/ and https://www.amiiboapi.com/?

We should not fake our User Agent to validate links. I think we should remove these links if they're not able to answer correctly when given a correct User Agent.

SitramSoft:
https://www.petfinder.com/developers/v2/docs/ and https://www.amiiboapi.com/ are the links I was refering to.

From my tests, I found out that the two websites return a 404 page when accessed with no user agent or a dummy one. When I used Chrome's user agent in the script, a valid page with the description of the API description is returned. 

I chose a user agent from an existing browser in the script since the use case I had in mind was not accessing the API directly but the documentation page. In this use case, the developer which is interested in this API will access the documentation page using a browser first. If the website can be accessed with a browser then from my point of view the link is valid.

It's up to you what you want to do with those two links.

yannbertrand:
I sent an email to Petfinder support and opened [an issue](https://github.com/N3evin/AmiiboAPI/issues/43) on AmiiboAPI. Will wait for them to fix the issue.

N3evin:
Seems to be a host issue. Sorry for the inconvenient. Is it possible to get delisted? I will request again once I decided to change host.

yannbertrand:
Sure thing. Thanks for looking at it. Let us know! 😄

PR #1542, opened 2021-01-29
Add the Hacker News API

eppeque:
I've added the Hacker News API in the README.md file. It's a great API, free to use and there is no API key required. Thank you for your attention.


eppeque:
Hacker News is already there. Not seen 😅

PR #1593, opened 2021-03-20
Add check for checking duplicated links in README file

### LINKED ISSUE #1592: Add extra validation to check if more than two links are the same
There should be a check in the validation to check if more than two links are the same. 
If they exist in the README the "merge" button should be blocked to merge the PR.

marekdano:
Add extra check into validate_links.py file to check if there are some duplicated links in the README file

I create an ignored list for some duplicated links which are not related to any API link. There is this link `https://github.com/davemachado/public-api` in the README listed twice. Should we ignore it too?

Closes #1592

### DIFF
--- a/build/validate_links.py
+++ b/build/validate_links.py
@@ -5,6 +5,11 @@ import re
 import socket
 import sys
 
+ignored_links = [
+    'https://github.com/public-apis/public-apis/actions?query=workflow%3A%22Run+tests%22', 
+    'https://github.com/public-apis/public-apis/workflows/Validate%20links/badge.svg?branch=master', 
+    'https://github.com/public-apis/public-apis/actions?query=workflow%3A%22Validate+links%22',
+]
 
 def parse_links(filename):
     """Returns a list of URLs from text file"""
@@ -16,6 +21,29 @@ def parse_links(filename):
     links = [raw_link[0] for raw_link in raw_links]
     return links
 
+def dup_links(links):
+    """Check for duplicated links"""
+    print(f'Checking for duplicated links...')
+    hasError = False
+    seen = {}
+    dupes = []
+
+    for x in links:
+        if x in ignored_links:
+            continue
+
+        if x not in seen:
+            seen[x] = 1
+        else:
+            if seen[x] == 1:
+                dupes.append(x)
+
+    if not dupes:
+        print(f"No duplicated links")
+    else:
+        print(f"Found duplicated links: {dupes}")  
+        hasError = True  
+    return hasError
 
 def validate_links(links):
     """Checks each entry in JSON file for live link"""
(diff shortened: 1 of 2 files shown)


matheusfelipeog:
Very good. I will review 😆 

> There is this link https://github.com/davemachado/public-api in the README listed twice. Should we ignore it too?

Yes, sounds reasonable to me.

build/validate_links.py
```diff
@@ -16,6 +21,29 @@ def parse_links(filename):
...
     return links
 
+def dup_links(links):
+    """Check for duplicated links"""
+    print(f'Checking for duplicated links...')
+    hasError = False
+    seen = {}
+    dupes = []
+
+    for x in links:
```
matheusfelipeog, line 31: Isn't it better to make it explicit?
```suggestion
    for link in links:
```

build/validate_links.py
```diff
@@ -16,6 +21,29 @@ def parse_links(filename):
...
+        if x in ignored_links:
+            continue
+
+        if x not in seen:
+            seen[x] = 1
+        else:
+            if seen[x] == 1:
+                dupes.append(x)
```
matheusfelipeog, line 39: Many URLs contain the `/` character at the end. This section does not consider them.

This is returning at the moment:
``` bash
Found duplicated links: ['https://github.com/davemachado/public-api']
```

If you consider the character `/` at the end:
``` bash
Found duplicated links: ['https://mailboxlayer.com', 'https://github.com/davemachado/public-api', 'https://languagelayer.com']
```

The `str.rstrip()` function solves this.

```suggestion
        x = x.rstrip('/')
        if x in ignored_links:
            continue

        if x not in seen:
            seen[x] = 1
        else:
            if seen[x] == 1:
                dupes.append(x)
```

matheusfelipeog (changes requested):

build/validate_links.py
marekdano: @matheusfelipeog Thanks for reviewing my PR and spotting an extra check with `/` character.

matheusfelipeog (approved):

matheusfelipeog:
Great 😄 

Now wait @yannbertrand review to merge.

yannbertrand:
I noticed this PR failed because there was a duplicate link detected 😅

It seems like there is already some validation done on https://github.com/public-apis/public-apis/blob/5dc97c529e0bdd23c03d5a1840445c9c2c356e2c/build/validate_format.py#L74

Is it possible to have only one verification rather than two? I'm not sure the old way did work as I don't remember seeing the error. It would be even better to have the list of all dupes you added in the logs. So I suggest to remove the old way.

yannbertrand:
> > There is this link https://github.com/davemachado/public-api in the README listed twice. Should we ignore it too?
> 
> Yes, sounds reasonable to me.

:+1:

marekdano:
> I noticed this PR failed because there was a duplicate link detected 😅
> 
> It seems like there is already some validation done on
> 
> https://github.com/public-apis/public-apis/blob/5dc97c529e0bdd23c03d5a1840445c9c2c356e2c/build/validate_format.py#L74
> 
> Is it possible to have only one verification rather than two? I'm not sure the old way did work as I don't remember seeing the error. It would be even better to have the list of all dupes you added in the logs. So I suggest to remove the old way.

Done @yannbertrand

yannbertrand (approved):
LGTM! Thank you

Issue #1590, opened 2021-03-19
The links returning 404 should be removed.


marekdano:
Here is a list of links that return 404

...

Please feel free to add more broken links when you find them.

marekdano:
Should we remove the broken link?

matheusfelipeog:
> Should we remove the broken links?

I believe so. I see no reason to keep them if they are broken. 🤔

yannbertrand:
Let's remove them 🙂

PawelBorkar:
Ofcourse.

PawelBorkar:
There's a PR regarding the link http://developer.linkedin.com/docs/rest-api . I had reviewed it and ask for the feedback of the author.

Anmoldeep1509:
Is the issue closed, if not, may I verify and remove the bad links?

marekdano:
The issue is still open.

PawelBorkar:
> Is the issue closed, if not, may I verify and remove the bad links?

Yeah! you can go ahead.

Anmoldeep1509:
API | STATUS | API STATUS
-- | -- | --
https://apility.io/apidocs/ | Already Removed | Down ❎
http://developer.linkedin.com/docs/rest-api | Already Removed | Down ❎
https://sentim-api.herokuapp.com/ | Already Removed | Up ✅
https://quarantine.country/coronavirus/api/ | Already Removed | Down ❎

FrankR-7:
I see the issue is still open. What if we made a script that periodically checks the links and alerts when any link is broken so we can manually remove it? (Or let it automatically remove the links)

marekdano:
@Anmoldeep1509 `https://quarantine.country/coronavirus/api/` is UP again.

Anmoldeep1509:
> https://quarantine.country/coronavirus/api/

yes, thanks, checked just now, it is working, updating the table, 

btw, do we have any md code to add beside the API - to show if it is up right or not?

Like I know a few repos where the md code checks if the website is up or not, similar will be needed for this case.

Anmoldeep1509:
> @Anmoldeep1509 `https://quarantine.country/coronavirus/api/` is UP again.

yes, thanks, checked just now, it is working, updating the table,

btw, do we have any md code to add beside the API - to show if it is up right or not?

Like I know a few repos where the md code checks if the website is up or not, similar will be needed for this case.

marekdano:
@yannbertrand do we have "MD code checks"?

[...]

Issue #1599, opened 2021-03-23
Links returning 403 should be removed


marekdano:
Here is a list of APIs that return 403

https://ipinfodb.com/api

vmbansal2001:
I want to contribute on this

dilbersha:
still on?

pawelborkar:
@dilbersha yup!

berkaysavas:
Good

matheusfelipeog:
I'm going to close this issue, this API is apparently working.

PR #1738, opened 2021-05-23
Remove broken links

iamjoker021:
- [x ] My submission is formatted according to the guidelines in the [contributing guide](CONTRIBUTING.md)
- [x ] My addition is ordered alphabetically
- [x ] My submission has a useful description
- [x ] The description does not end with punctuation
- [x ] Each table column is padded with one space on either side
- [x ] I have searched the repository for any relevant issues or pull requests
- [ ] Any category I am creating has the minimum requirement of 3 items
- [ ] All changes have been [squashed][squash-link] into a single commit

[squash-link]: <https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit>

| Problem | Broken links in repository (Issue #1590 and #1590)|
|---|---|
| Problem Description | There are about 800 api links among them there are some broken links need to be removed |
| Solution | Instead of manually checking the them I suggest a program(bad_link_filter.py) that would save bad links in separate file so that can be reviewed later |

The program successfully narrowed 800 links to 32 links (Executed on 23-05-2020)

However some of links works return error because of bot security feature in respective webpage

Hence a manual test only on those filtered records in needed

I've done the manual check also and saved them in broken-links.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,11 +230,8 @@ API | Description | Auth | HTTPS | CORS |
 | [CryptoMarket](https://developers.cryptomkt.com) | Cryptocurrencies Trading platform | `apiKey` | Yes | Yes |
 | [Cryptonator](https://www.cryptonator.com/api/) | Cryptocurrencies Exchange Rates | No | Yes | Unknown |
 | [Gemini](https://docs.gemini.com/rest-api/) | Cryptocurrencies Exchange | No | Yes | Unknown |
-| [ICObench](https://icobench.com/developers) | Various information on listing, ratings, stats, and more | `apiKey` | Yes | Unknown |
-| [MercadoBitcoin](https://www.mercadobitcoin.net/api-doc/) | Brazilian Cryptocurrency Information | No | Yes | Unknown |
 | [Nexchange](https://nexchange2.docs.apiary.io/) | Automated cryptocurrency exchange service | No | No | Yes |
 | [Poloniex](https://poloniex.com/support/api/) | US based digital asset exchange | `apiKey` | Yes | Unknown |
-| [Technical Analysis](https://technical-analysis-api.com) | Cryptocurrency prices and technical analysis | `apiKey` | Yes | No |
 | [VALR](https://docs.valr.com/) | Cryptocurrency Exchange based in South Africa | `apiKey` | Yes | Unknown |
 | [WorldCoinIndex](https://www.worldcoinindex.com/apiservice) | Cryptocurrencies Prices | `apiKey` | Yes | Unknown |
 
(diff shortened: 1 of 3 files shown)


pawelbr:
Hey @iamjoker021 
Thank you so much for making this PR. We appreciate it.

Since its a PR regarding enhancement. All the collaborators need to approve it. So I'll be requesting their reviews as well while giving my own.

iamjoker021:
Sorry I'm new to GitHub could you kindly elaborate what checks failed so that I could correct them. Thanks in advance!

matheusfelipeog (commented):
Hi @iamjoker021 😄 

Thanks for your contribution, but I believe that confusion has occurred here.

```
Instead of manually checking the them I suggest a program(bad_link_filter.py) that would save bad links in separate file so that can be reviewed later
```

We, collaborators, do not do manual verification on each of the more than 800 links listed here. We have the [validate_links.py](https://github.com/public-apis/public-apis/blob/master/build/validate_links.py) script that does this job for us, which shows us only the links that may need manual review. This script is executed by [Github Actions](https://docs.github.com/en/actions) and shown the result to us in the [Actions panel](https://github.com/public-apis/public-apis/actions).

An example of this process is here:

image

After manually checking the links returned by the _validate_links.py_ script, we open issues listing the links that need to be deleted or updated so that others can contribute to the project.

Examples of these problems are: #1590 and #1599

The validate_links.py script, in addition to checking for broken links, also checks for links that need to be updated due to redirection (HTTP code> 300 and <399).

Therefore, this pull request is not much different from the _validate_links.py_ script, which, in essence, does the same thing. One of the only differences is to save the broken links in a file, but I don't see any real gain in keeping them since they are broken.

I will wait for the other collaborators to see if there is a different analysis of this Pull Request.

marekdano (commented):
@iamjoker021 thanks a mill for your PR and for considering how we can improve the flow of finding the broken links and removing them.

@matheusfelipeog thanks for well explaining the current flow that we have for finding the broken links and how we remove them! 

I think we don't need to have another script for checking the broken links. What if we can improve the [validate_links.py](https://github.com/public-apis/public-apis/blob/master/build/validate_links.py) script? 
I also think that we don't need to have another file for listing all broken links. We can find out the links that return something different than 200 in the [Github Action](https://github.com/public-apis/public-apis/runs/2661825158?check_suite_focus=true) as @matheusfelipeog said above.

What if we add to automatically remove the links which return 404 and we don't want to check them manually before?

matheusfelipeog:
> What if we add to automatically remove the links which return 404 and we don't want to check them manually before?

I particularly prefer to do a manual check in these cases. So I can check if the link is broken due to the discontinuation of the service or change in the address of the page. So in case of discontinuation I can delete it and in case of changing the address of the page I can create a issue informing that the url needs to be updated. 🤔

matheusfelipeog:
The PR author deleted the fork and it looks like we've already made a decision here.

Fechando este PR.

PR #1749, opened 2021-05-27
New nameday API url

xnekv03:


pawelborkar (approved):
Thanks for the contribution @xnekv03

PR #1842, opened 2021-07-12
Removed gofile.io, file download is not working

kamal-hossain:
## Reasons for removing this site
- Download link they are providing is not working.
- They are not providing any direct link to the user who is not a donor.


fintzd (approved):
Thanks!

marekdano:
I checked their service and the download works for me.

Can someone else have an issue with the `gofile.io` service too?

pawelborkar:
Yeah, It's working for me as well.

matheusfelipeog:
For me it's also working

matheusfelipeog:
I'm going to close this PR. Looks to me like it was just a temporary problem.

If other people have this kind of problem, please let us know ;)

PR #1973, opened 2021-08-21
Updated Smartsheet link to current, old link deprecated

julschong:
- [x] My submission is formatted according to the guidelines in the [contributing guide](/CONTRIBUTING.md)
- [x] My addition is ordered alphabetically
- [x] My submission has a useful description
- [x] The description does not end with punctuation
- [x] Each table column is padded with one space on either side
- [x] I have searched the repository for any relevant issues or pull requests
- [x] Any category I am creating has the minimum requirement of 3 items
- [x] All changes have been [squashed][squash-link] into a single commit

[squash-link]: <https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit>

PR #1974, opened 2021-08-21
Revert "Updated Smartsheet link to current, old link deprecated"

soffieprincess09:
Reverts public-apis/public-apis#1973


pawelborkar:
Why do you want to revert this?

PR #2030, opened 2021-09-05
Remove CoinDesk API

robvanbakel:
The [CoinDesk](http://www.coindesk.com/api/) link was present, but pointed to a 404-page. The endpoint seems to still be working ([example](https://api.coindesk.com/v1/bpi/currentprice.json)), but is nowhere documented or mentioned by CoinDesk, hence the removal from this list.

Alternatively, we could update the CoinDesk API link to be this one endpoint, without any further context or documentation available.
```
| [CoinDesk](https://api.coindesk.com/v1/bpi/currentprice.json) | Bitcoin Price Index | No | No | Unknown |
```


matheusfelipeog (approved):
Thanks for removing this API and checking if there was a link to the new documentation. <3

I believe it really is better to remove it in that case.

PR #2066, opened 2021-09-24
Remove SWAPI duplication

boversoneg:
just removing duplication of one api


matheusfelipeog (commented):
Thanks for your contribution @boversoneg 😄 

I understand, they really are very similar. However, they are not the same API.

Both are forks of the [same github project](https://github.com/phalt/swapi), but contain different data to query. So I don't think it's necessary to remove one of them from the listing.

PR #2073, opened 2021-09-30
Updating the CORS value for restcountries.com

amatosg:
Is just an update :)


marekdano:
Thanks for your contribution, @amatosg!

PR #2947, opened 2021-12-11
Changed redirect limit in the file.

codewithpom:
I changed the redirect limit and I want to know that why do we not use requests or grequests as they are well developed

### DIFF
--- a/build/validate_links.py
+++ b/build/validate_links.py
@@ -49,7 +49,7 @@ def validate_links(links):
     print(f'Validating {len(links)} links...')
     hasError = False
     for link in links:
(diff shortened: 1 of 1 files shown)


codewithpom:
@matheusfelipeog as I said I am here 😎

matheusfelipeog (commented):
Thanks for your new contribution @codewithpom, but this change is not necessary. We use this timeout to generate an exception if the site is slow to respond.

> I want to know that why do we not use requests or grequests as they are well developed

That will happen soon. I'll improve our scripts and test flow in one of their upcoming updates.

codewithpom:
Ok closing it 😭

PR #2957, opened 2021-12-18
Updated MeowFacts

codewithpom:
Change cors for cats API.


matheusfelipeog (approved):
Thanks for the update @codewithpom ;)

PR #2968, opened 2021-12-22
Changed cors for urlHaus

codewithpom:
Changed cors for urlHaus to yes


pawelborkar (approved):
Thanks for the update @codewithpom

PR #2976, opened 2021-12-25
Upd Brazilian Vehicles and Prices: CORS from Unknown to No

codewithpom:
Updated cors to no


codewithpom:
Merry Christmas

matheusfelipeog (approved):
Thanks for your new update @codewithpom 🎅

PR #2978, opened 2021-12-27
Add Churchs Data API

Uolary:
Add Churchs Data API to Geocoding


matheusfelipeog (commented):
Thanks for your contribution @Uolary, but the documentation for this API is apparently not adequate. 

It just documents the endpoints to be used, but it doesn't tell you their limits or where you can get an API key, for example. Therefore, this API cannot be listed here.

PR #2993, opened 2022-01-06
Remove Troposphere

ManuelBauer:
Removed Troposphere weather API, because the service is no longer active.


matheusfelipeog (approved):
Oh right @ManuelBauer. Thanks  for removing it ;)

PR #3397, opened 2022-12-28
Changed link for LocationIQ

sagar-locationiq:
From locationiq.org to locationiq.com
Corrected description


Gojekgobiz:
> From locationiq.org to locationiq.com Corrected description

PR #3410, opened 2023-01-10
Update README.md

samyakjain1908:
AniApi link was changed as the old link was not working. 

Error : Page not found


CaedenPH (approved):
Nice catch

PR #3531, opened 2023-06-09
Remove CountAPI [DEPRECATED]

alexlostorto:
Count API went offline a couple of weeks ago.


JasonLovesDoggo:
I've had my own completely open sourced version hosted for around half a year now, It's a mostly drop-in solution  http://abacus.jasoncameron.dev/ https://github.com/JasonLovesDoggo/abacus

PR #3709, opened 2023-11-26
Removed FOAAS as it seems down

Nereuxofficial:
I removed FOAAS as it seems to no longer exists and just points to an empty heroku page


README.md
peterthengai, line 1: Line 1507 should also be removed as it also references the Fuck Off As A Service

Nereuxofficial: Good catch! 
I added the change to the PR!

PR #4391, opened 2025-06-25
Add Openrouter API

Alorse:
Adds Openrouter API to new Artificial Intelligence category


Alorse:
> Does this API require a token? I got redirected to a login page while testing.

It requires an `apiKey`, I added the parameter into the `Auth` column

PR #4427, opened 2025-06-30
Add Bored API to Games & Comics section

prashanshasharma:
Added Bored API which returns fun activity suggestions to beat boredom. No Auth required, supports HTTPS and CORS.


eka0789:
**_Nice addition!_** The Bored API is fun and practical. No auth and CORS support make it easy to integrate. Would be great to add a sample response for quick preview. Good job! 👏

PR #4984, opened 2025-11-16
Add CountryLayer API

2705DeeptiN:
Added the CountryLayer API under the Geocoding category.

CountryLayer provides detailed country information such as population, borders, capital city, languages, currencies, geographic coordinates, regions, and ISO codes. The API requires an API key and supports HTTPS.


Salozar73:
Hello. This repository is no longer active. You can post on https://api-hub.pro/ if you are not already there.

PR #5202, opened 2026-02-05
Add Moltpost

AlephArcher:
Moltpost (https://moltpost.io) is a REST API for sending physical postcards worldwide. No authentication required, no signup needed. The API endpoint is freely accessible — payment is handled via Stripe checkout links in the response.


Salozar73:
Hello. This repository is no longer active. You can post on https://api-hub.pro if you are not already there.

PR #5461, opened 2026-03-14
Add market lens api

yoghesh:
Add market lens api to finance

closes issue #5460


BurnyCoder:
I take it back, don't add it.

PR #5528, opened 2026-03-20
Add API Status Check

shibley:
API Status Check provides a free, public JSON API for real-time status monitoring of 270+ third-party APIs including OpenAI, AWS, Stripe, GitHub, Vercel, Cloudflare, and more.

**API endpoint:** `GET https://apistatuscheck.com/api/status`

Returns JSON with current operational status for all monitored APIs. No authentication required, HTTPS supported.

Example response:
```json
{
  "count": 272,
  "lastUpdated": "2026-03-20T18:36:05Z",
  "apis": [
    {
      "slug": "chatgpt",
      "name": "ChatGPT",
      "category": "ai",
      "status": "operational"
    }
  ]
}
```

Individual API status: `GET /api/badge/{slug}` (returns SVG badge)

Placed in Development section alphabetically between API Grátis and ApicAgent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -481,6 +481,7 @@ API | Description | Auth | HTTPS | CORS |
 | [24 Pull Requests](https://24pullrequests.com/api) | Project to promote open source collaboration during December | No | Yes | Yes |
 | [Agify.io](https://agify.io) | Estimates the age from a first name | No | Yes | Yes |
 | [API Grátis](https://apigratis.com.br/) | Multiples services and public APIs | No | Yes | Unknown |
+| [API Status Check](https://apistatuscheck.com/api/status) | Real-time status monitoring for 270+ third-party APIs | No | Yes | Unknown |
 | [ApicAgent](https://www.apicagent.com) | Extract device details from user-agent string | No | Yes | Yes |
 | [ApiFlash](https://apiflash.com/) | Chrome based screenshot API for developers | `apiKey` | Yes | Unknown |
 | [APIs.guru](https://apis.guru/api-doc/) | Wikipedia for Web APIs, OpenAPI/Swagger specs for public APIs | No | Yes | Unknown |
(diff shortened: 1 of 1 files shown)


shibley:
Hi! Just a friendly follow-up on this PR. API Status Check provides a free public API (`/api/status`) that returns real-time status for 500+ major APIs — a natural fit alongside the other monitoring/status entries in the list. Happy to make any adjustments needed. Thanks for maintaining such an incredible resource!

shibley:
Following up on this after a few months, with two corrections pushed rather than just a bump:

1. **Count corrected.** The entry said "270+ third-party APIs"; the endpoint currently reports exactly **285** (`count` field in `https://apistatuscheck.com/api/status`). Updated to the real number so the entry isn't a claim that drifts.
2. **CORS column corrected — twice.** I first set it to `Yes`, then actually checked the response headers, found no `Access-Control-Allow-Origin`, and set it back to `No`. That was accurate at the time of writing.

Worth noting for whoever reviews: CORS has now been implemented on that endpoint (`Access-Control-Allow-Origin: *` plus an `OPTIONS` preflight handler) and is queued for deploy. I've deliberately left the column at `No` until it's live — I'll push one more commit flipping it to `Yes` once you can verify it with `curl -I`, rather than asking you to merge a claim that isn't true yet.

`Auth: No`, `HTTPS: Yes` are both still accurate. Entry is one table row in the Development section, alphabetically placed, and mergeable against current `master`.

shibley:
Following up with a fix rather than a bump: the CORS column on this entry now says `Yes`, and it's verifiable right now.

When I opened this PR I set CORS to `No` because the endpoint genuinely didn't send the header. I've since added `Access-Control-Allow-Origin: *` plus an `OPTIONS` preflight handler across every response path (200 / single-API / 404 / rate-limited), deployed it, and only then flipped the column — so the table matches reality instead of the other way round:

```
$ curl -sI https://apistatuscheck.com/api/status | grep -i access-control
access-control-allow-methods: GET, OPTIONS
access-control-allow-headers: Content-Type, x-api-key
access-control-max-age: 86400
```

The service count in the description is also live-checked against the endpoint itself (`count: 285` as of today), not a marketing number — I corrected it down from an inflated figure earlier and will keep it honest if it drifts.

The current diff is a single line in the Development section, alphabetically placed. Happy to reword the description, move it to a different section, or close this out if it isn't a fit — just say which.

PR #5958, opened 2026-04-26
Update Currents API description

Autumn-Atlas:
Update the Currents API description to better reflect its features: real-time and historical global news with multilingual support.


ArindamOSS (approved):
Thanks for contributing!

PR #5976, opened 2026-04-30
Add Satoshi API

Bortlesboat:
Adds Satoshi API to the Cryptocurrency section.\n\nVerification:\n- Docs: https://bitcoinsapi.com/docs\n- Anonymous request succeeds: https://bitcoinsapi.com/api/v1/fees/recommended returns HTTP 200 without an API key\n- HTTPS supported\n- CORS headers are not emitted for arbitrary web origins, so the CORS column is set to No


Bortlesboat:
Closing because the listing is no longer valid for this catalog: both https://bitcoinsapi.com/docs and the free /api/v1/fees/recommended endpoint currently return Cloudflare Tunnel error 1033 (HTTP 530). I won't refresh a public API entry while its documentation and unauthenticated endpoint are unavailable.

PR #5983, opened 2026-05-02
Remove Age of Empires 2 API as it is Invalid

notashwin0:
In the Games and Comics section, upon opening the Age of Empires 2 API, the website seems to be down.


ArindamOSS (approved):
Thanks for Pointing this out.

PR #6003, opened 2026-05-04
Add FileUp API to Cloud Storage & File Sharing

intagliated:
## Pull Request Checklist

- [x] I have read and followed the [contribution guidelines](https://github.com/public-apis/public-apis/blob/master/CONTRIBUTING.md)
- [x] I have read and followed the [API addition checklist](https://github.com/public-apis/public-apis/blob/master/CONTRIBUTING.md#api-addition-checklist)
- [x] I confirm this API has a free tier or is completely free
- [x] I have added a single API entry per pull request
- [x] I have verified the API name does not end with "API" or contain a TLD
- [x] I have verified the description is under 100 characters and has no ending punctuation
- [x] I have verified the entry is alphabetically ordered within its category
- [x] I have squashed all my commits into a single commit

## Additional Information

**API Documentation:** https://github.com/RealSinaSnp/FileUp

**Test Request:**

curl -L -F "file=@test.txt" -F "expireMinutes=30" -F "maxViews=5" https://upload.sinasnp.com/api/public/upload

**Test Response:**

{"fileName":"test_20260504043303.txt","size":38,"url":"http://upload.sinasnp.com/files/public/txt/test_20260504043303.txt","expireAt":"2026-05-04T04:33:03.973346Z","maxViews":5}

**What this API provides:**
- Temporary file hosting with no authentication required
- Configurable expiration times (expireMinutes parameter)
- Configurable view limits (maxViews parameter)
- Returns file URL, size, expiration time, and max views
- Free tier with 100MB max file size

**Source:** FileUp - Open source temporary file hosting service

PR #6065, opened 2026-05-11
Fix format validator category parsing

### LINKED ISSUE #6041: [BUG] validator crashes with UnboundLocalError
The validator script crashes with an `UnboundLocalError` when running format validation on `README.md`. This prevents valid pull requests from passing the automated checks.

### Error Log

When running `python scripts/validate/format.py README.md`, the following errors occur:

SyntaxWarning: "\s" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\s"? A raw string is also an option.
SyntaxWarning: "*" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\*"? A raw string is also an option.
SyntaxWarning: "[" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\["? A raw string is also an option.
Traceback (most recent call last):
File "scripts/validate/format.py", line 277, in <module>
main(filename)
File "scripts/validate/format.py", line 259, in main
file_format_err_msgs = check_file_format(lines)
File "scripts/validate/format.py", line 197, in check_file_format
alphabetical_err_msgs = check_alphabetical_order(lines)
File "scripts/validate/format.py", line 74, in check_alphabetical_order
categories, category_line_num = get_categories_content(contents=lines)
File "scripts/validate/format.py", line 65, in get_categories_content
categories[category].append(title)
UnboundLocalError: cannot access local variable 'category' where it is not associated with a value

### Possible Cause

The `get_categories_content` function attempts to reference the variable `category` in the `except` block before it is assigned a value in the `try` block. This causes the script to crash when it encounters a specific line pattern (likely blank lines or horizontal rules `|---|`).

Additionally, the regular expressions contain invalid escape sequences (`\s`, `\*`, `\[`) that should be raw strings (e.g., `r'\s(.+)'`).

### Example : PR Affected

This bug is blocking PR #6003: [Add FileUp API to Cloud Storage & File Sharing](https://github.com/public-apis/public-apis/pull/6003)

### Expected Behavior

The validation script should either:
1. Successfully validate correctly formatted API entries, or
2. Provide clear error messages about actual formatting issues

### Environment

- Python version: 3.14
- Script location: `scripts/validate/format.py`

### Possible Solution

1. Fix the regex patterns by using raw strings (add `r` prefix)
2. Initialize the `category` variable before the try-except block in `get_categories_content()`

Example fix:
```python
category = None  # Initialize before try block
try:
    # existing code
except Exception:
    if category is not None:
        # handle error
```
### Additional Context
This issue was discovered while attempting to contribute a new API entry. The entry itself follows all formatting guidelines, but the validation script fails due to a code bug rather than any formatting error.

MukundaKatta:
## Summary
- prevent the format validator from reading table rows before the first API category
- skip Markdown table delimiter/header rows consistently
- use raw regex strings to avoid invalid escape sequence warnings

Fixes #6041.

## Verification
- python3 -m py_compile scripts/validate/format.py
- python3 scripts/validate/format.py README.md (now completes and reports validation errors instead of crashing with UnboundLocalError)
- in-memory regression fixture for pre-category promo tables and category API tables

### DIFF
--- a/scripts/validate/format.py
+++ b/scripts/validate/format.py
@@ -3,7 +3,7 @@
 import re
 import sys
 from string import punctuation
-from typing import List, Tuple, Dict
+from typing import Dict, List, Optional, Tuple
 
 # Temporary replacement
 # The descriptions that contain () at the end must adapt to the new policy later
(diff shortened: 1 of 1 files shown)


intagliated:
Hi, 
I tested and confirmed  the working of your script and  it  fixes #6041
The validator no longer crashes and runs to completion.
This fix  also immediately unblocks PR #6003 (Add FileUp API) which was failing CI due to this crash.
Now that the validator runs, I'm seeing other pre-existing validation issues (false positives, case sensitivity problems, column counting issues). These aren't caused by your PR but it would be great if you could do some follow-up contributions.

Ready for maintainer review.

intagliated:
@ArindamOSS  do check @MukundaKatta's PR  to the issue #6041 and see if the validator script's fix  can be incorporated into the repository.

PR #6208, opened 2026-06-01
Remove Zoo Animals API

Lysine000:
The Zoo Animals API returns a 404 error from the Heroku router, indicating it is no longer active. Following the contributing guidelines, I am removing this dead link.


ArindamOSS (approved):
Thanks for pointing this out to us and making this PR

PR #6348, opened 2026-06-18
Update InfraNode API

street1983nk:
Corrects the Auth field for the existing **InfraNode** entry in the Open Data section.

InfraNode is fully **keyless** (no API key, no signup), so the Auth column should be `No`, not `apiKey`. This can be verified by calling any endpoint without credentials, e.g. `https://infranode.dev/api/v1/cities/koeln/weather`.

Only the Auth field of the single existing line is changed; formatting and alphabetical order are untouched.


street1983nk:
Heads-up for reviewers: the failing **Validate README.md changes** check is not caused by this entry. The diff is a one-line change to an existing row (correcting InfraNode's Auth from `apiKey` to `No`, since the API is keyless), placed alphabetically and matching the column format.

The check fails with an `UnboundLocalError: local variable 'category' referenced before assignment` raised inside `scripts/validate/format.py` (`get_categories_content`, line 65) — i.e. the validator itself errors out while parsing the README, rather than reporting a formatting problem with this row. The `Run tests of validate package` check passes.

Happy to adjust anything if needed. Thanks for maintaining the list!

xiaolaobantech:

PR #6368, opened 2026-06-20
Fix format.py crash on Python 3.13+

OmniaZ1:
## Problem

format.py crashes on Python 3.10+ with `UnboundLocalError` when the README contains markdown table rows before the first `###` category header.

**Root cause:** `get_categories_content()` uses `category` without initializing it. When table rows with `[TITLE](LINK)` appear before any `###` anchor (like the APILayer ad table), the code tries `categories[category].append(title)` where `category` was never assigned.

**Secondary issue:** Three regex patterns use bare strings with backslash escapes instead of raw strings, producing `SyntaxWarning` on Python 3.12+ (will become `SyntaxError` in future Python).

## Fix

1. Initialize `category = None` and add guard `if category is None: continue`
2. Convert regex patterns to raw strings

## Testing

- All 31 existing unit tests pass unchanged
- format.py now runs successfully on the full README.md

### DIFF
--- a/scripts/validate/format.py
+++ b/scripts/validate/format.py
@@ -24,9 +24,9 @@ num_segments = 5
 min_entries_per_category = 3
 max_description_length = 100
 
-anchor_re = re.compile(anchor + '\s(.+)')
-category_title_in_index_re = re.compile('\*\s\[(.*)\]')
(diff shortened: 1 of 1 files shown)


xiaolaobantech:

PR #6406, opened 2026-06-25
Add ReefAPI

b21946735:
Adds **ReefAPI** to the Development category (alphabetical, between Random Stuff and Rejax).

ReefAPI returns live web data from 160+ sources (search, social, e-commerce, real estate, jobs, travel, news, finance, company/domain/people intelligence) as clean JSON over one endpoint group. It has a free tier (1,000 credits, no card required), same class as the existing ScraperApi / ScrapeNinja entries in this section.

- API: `apiKey` · HTTPS: Yes · CORS: Unknown
- Site: https://reefapi.com · Docs: https://reefapi.com/docs

Single-line addition, alphabetical placement, follows the table format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -598,6 +598,7 @@ API | Description | Auth | HTTPS | CORS |
 | [Qrcode Monkey](https://www.qrcode-monkey.com/qr-code-api-with-logo/) | Integrate custom and unique looking QR codes into your system or workflow | No | Yes | Unknown |
 | [QuickChart](https://quickchart.io/) | Generate chart and graph images | No | Yes | Yes |
 | [Random Stuff](https://api-docs.pgamerx.com/) | Can be used to get AI Response, jokes, memes, and much more at lightning-fast speed | `apiKey` | Yes | Yes |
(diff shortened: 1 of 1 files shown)


b21946735:
Friendly bump - happy to rebase or adjust the entry if anything needs changing. The API is still live and the free tier is unchanged. Thanks for maintaining this list.

b21946735:
Two corrections to the entry in this PR, both mine:

- **`160+` should be `180+`.** The count grew since June and it moves, so a floor is the honest number rather than an exact one that goes stale again.
- **CORS should be `No`, not `Unknown`.** I measured it rather than guessing: a preflight `OPTIONS` returns 405 and no `Access-Control-Allow-*` header is sent on a cross-origin `GET`. That is deliberate on our side, since the API takes a key and we would rather it not be called straight from a browser.

So the line should read:

```
| [ReefAPI](https://reefapi.com) | Live web data from 180+ sources (search, social, e-commerce, real estate, jobs, finance) returned as clean JSON | `apiKey` | Yes | No |
```

Happy to push the corrected commit if you would prefer the branch updated instead of editing on merge. The API and its free tier are unchanged.

PR #6444, opened 2026-06-29
Correct Road511 CORS to No

travelcamnet:
The Road511 entry lists CORS as `Yes`, but the API only allows an allowlist of its own origins (not a wildcard). A cross-origin request from an arbitrary site returns no `Access-Control-Allow-Origin` header, so browser-based cross-origin access is not supported.

This one-cell change updates the CORS column from `Yes` to `No` to match actual behavior. No other fields changed.


daltino (commented):
**DECISION: APPROVE**

This PR makes a simple and consistent change to the `CORS` field for the Road511 API entry, updating it from `Unknown` to `No`. The modification aligns with the contribution guidelines, providing accurate information about the API's CORS support. Everything looks good to merge!

PR #6460, opened 2026-07-01
Add SSRF guard to the link validator

nate-avra:
> Note: this is a security fix to the repository's CI tooling (`scripts/`), **not** an API list addition — so the list-entry checklist in the PR template does not apply here.

## What

Add scheme + private-IP validation to the link checker (`scripts/validate/links.py`) so it refuses to request non-`http(s)` URLs or hosts that resolve to private / reserved / loopback / link-local addresses.

## Why

`check_if_link_is_working()` passes each README URL straight to `requests.get()`. Because anyone can open a PR adding an entry, a URL such as `http://169.254.169.254/latest/meta-data/` or `http://127.0.0.1/` would make the CI runner issue a request to an internal address — a server-side request forgery (SSRF) vector.

This is defense-in-depth hardening of the CI tooling: it removes the obvious SSRF surface (internal IPs and non-HTTP schemes). It is not a full DNS-rebinding mitigation, since `requests.get()` re-resolves the host afterward — but it closes the direct case reachable via a pull request.

## Change

- `validate_url()` rejects unsupported URL schemes and blocks hostnames whose literal or resolved IPs are private/reserved/loopback/link-local, before any request is made.
- `check_if_link_is_working()` calls it first and short-circuits with an `ERR:URL:` message.
- No behavior change for the legitimate public `https` API links already in the list.

## Testing

- Added cases in `scripts/tests/test_validate_links.py`: blocked schemes, blocked private/reserved IPs (including cloud-metadata `169.254.169.254`), allowed public URLs, and `is_private_ip`. They use literal IPs, so they require no network access.
- Full suite green — `cd scripts && python -m unittest discover tests/`:

```
Ran 35 tests in 0.002s

OK
```

### DIFF
--- a/scripts/tests/test_validate_links.py
+++ b/scripts/tests/test_validate_links.py
@@ -7,6 +7,8 @@ from validate.links import check_duplicate_links
 from validate.links import fake_user_agent
 from validate.links import get_host_from_link
 from validate.links import has_cloudflare_protection
+from validate.links import validate_url
+from validate.links import is_private_ip
 
 
 class FakeResponse():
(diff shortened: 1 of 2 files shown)


xiaolaobantech:
Built a MiniMax relay for CN users.
$3/1M tokens base. OpenAI-compatible. USDT.
xiaolaoban12345.top

yyy708:
Compatible with MiniMax, DeepSeek, GLM, Qwen and other mainstream models
Shop now: yyyyybl.top

xiaolaobantech:

PR #6466, opened 2026-07-02
Add Bargo Congress Trades API

bargo-ai:
Adds Bargo Congress Trades API to the public API list.

Free tier limits from the docs:
- Keyless: 30 requests/day and 1,000 rows/day per IP
- Free API key: 1,000 requests/day and 25,000 rows/day

No credit card required.


xiaolaobantech:

PR #6482, opened 2026-07-04
Add Bolsai API

viniciuslazzari:
Adds Bolsai, a REST API for Brazilian stock market data (fundamentals, prices, dividends, and FII data), sourced from B3, CVM, and BCB. Free tier available.

- Docs: https://usebolsai.com/docs
- Added under Finance, alphabetically between Boleto.Cloud and BriefTape.


jasonguan1125-ux:
🔥 Multi-Model API Keys – Best Value Deal!

Looking for affordable AI API keys? You’re in the right place!

✅ New users get 1,000,000 tokens for $1 on their first purchase
✅ Supports MiniMax, DeepSeek, GLM/Qwen, and more popular models
✅ Prices as low as $3 per 1,000,000 tokens
✅ Instant delivery
✅ Secure and stable connection

💰 Same capabilities, just a fraction of the cost—don’t overpay!

🛒 Buy now: www.jasonapi.top

PR #6496, opened 2026-07-06
Correct InfraNode Auth: No (keyless), was apiKey

street1983nk:
One-line correction to the existing **InfraNode** entry under *Open Data*.

InfraNode is a keyless public API (no API key, no account). The entry currently lists Auth as `apiKey`, which is incorrect. Corrected to `No`.

- [x] Single existing entry corrected, no new entry
- [x] Description unchanged, under 100 chars, no trailing period
- [x] Link works and uses HTTPS
- [x] Format matches the guidelines


xiaolaobantech:

PR #6499, opened 2026-07-06
Add DNSai API

DNSai-app:
DNSai adds a few things the existing lookup tools in this section don't cover: DKIM selector discovery (finds a domain's signing keys without knowing selector names), SPF analysis with the 10-DNS-lookup limit check, plain-English interpretation of results aimed at non-experts, and a free API plus MCP server so the same checks run inside AI assistants. All basic lookups are free with no signup or ads.

Disclosure: I'm the founder of DNSai. Happy to adjust wording or placement, and no hard feelings if it's not a fit for the list.


xiaolaobantech:

xiaolaobantech:

PR #6503, opened 2026-07-07
Update Goldprice.dev description

michaelmustopo:
The Goldprice.dev entry's description was stale: it said "13 currencies" (actually 31) and didn't mention two products the API has since added (per-country physical dealer prices, and an options/prediction-market probability surface). This PR updates the description text only — name, link, Auth, HTTPS, and CORS columns are unchanged.


xiaolaobantech:

xiaren518:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at:

 jsz186910.top

PR #6651, opened 2026-07-22
Add OrbitalWiki API

Anomo1:
Adds OrbitalWiki to Science & Math.

OrbitalWiki is a free, public satellite catalog (16,000+ tracked objects) that merges CelesTrak orbital elements, Jonathan McDowell's GCAT launch records, and Wikidata enrichment into one catalog with per-field source provenance. Read-only REST API, free tier 1,000 req/day, key from the dashboard at https://orbitalwiki.com/developers.

- Alphabetized under Science & Math (between OpenAlex and Purple Air)
- Description is 94 characters (under the 100 char limit)
- One link, single PR per guidelines


daltino (changes requested):
The addition of the OrbitalWiki API is intriguing; however, the entry's description ends abruptly (“merging CelesTrak, GCAT”) and lacks clarity. It should explicitly state if the API is entirely free or includes a free tier, as per the contribution guidelines. Additionally, confirm whether it supports CORS to avoid an "Unknown" entry.

Anomo1:
Thanks for the review. Updated:

- Description now states the free tier explicitly: "free tier included"
- CORS confirmed by testing a live preflight request against the API (`Access-Control-Allow-Origin: *`), changed from Unknown to Yes

PR #6655, opened 2026-07-23
Add Dino.markets API

michaelmustopo:
Adds Dino.markets to Finance.

Matched Kalshi and Polymarket prediction-market data — the same event tracked across both venues, served as one feed. Covers sports, crypto, weather, and economics.

- Free tier available; most endpoints require an API key (`apiKey`).
- No CORS headers on requests (verified via a live GET against the API).
- Docs verified reachable: https://dino.markets/docs

Disclosure: I'm affiliated with this API.

Placed alphabetically between CongressInvests and Econdb, per the section's existing order.


michaelmustopo:
FYI on the failing "Validate README.md changes" check — traced it down, and it's not related to this entry. `scripts/validate/format.py` currently throws an unhandled `UnboundLocalError` in `get_categories_content()` on a completely clean, unmodified `master` (confirmed by running it directly against upstream `master` at `aacdeff`, no changes at all). The "APIs Covered Under APILayer Suite!" heading added to the top of the README isn't registered in the Index section the way the script's category parser expects, and it crashes before it reaches any real validation. Every PR against this repo right now will fail the same "Validate Markdown format" step for the same reason, independent of what's actually being added.

Not something fixable from this PR's diff — flagging in case it's useful for triage.

PR #6666, opened 2026-07-25
Add open-banking.io to Finance section

john-frandsen:
## Description

Added [open-banking.io](https://open-banking.io/) to the Finance section.

open-banking.io is a European open banking API that provides account information (AIS) and payment initiation (PIS) across 5,000+ EU banks. Its key differentiator is that it operates without requiring customers to purchase eIDAS/QWAC certificates (which typically cost €2,000–10,000/year), making PSD2 bank data access significantly more accessible to developers and SMBs.

- **Auth:** API key
- **HTTPS:** Yes
- **CORS:** Unknown
- Placed alphabetically between Nordigen and OpenFIGI


ArindamOSS (commented):
I don't see any Free Plan. Can you please cross-check?

tzxqjai:
Cool addition! It’s always great to see more regional public APIs available.
Speaking of APIs, many developers I work with are still struggling with expensive official LLM endpoints, region locks and complicated KYC requirements.
If anyone here is looking for stable, cost-effective OpenAI-compatible LLM APIs for their projects, feel free to hit me up via DM.

john-frandsen:
Thanks for checking! You're right to flag this — the README entry shouldn't imply a free tier where there isn't a clear one. 

To clarify: open-banking.io starts at ~€3/month (pay-as-you-go), not a traditional free plan. I'll update the entry to reflect the pricing model more accurately. The "Yes" in the Auth column refers to apiKey-based authentication, not a free tier.

Happy to update the entry if the maintainers prefer a different pricing column value.

PR #6689, opened 2026-07-27
Add TranscriptAPI API

scheemunai:


scheemunai:
Confirming that TranscriptAPI has a free tier: new accounts receive 100 free credits at signup, with no card required. The listed link points directly to the API documentation at https://transcriptapi.com/docs.

PR #6692, opened 2026-07-27
Add TranscriptFetch API

Transcript-Fetch-Admin:
Adds TranscriptFetch to the Video section.

Returns transcripts for YouTube, TikTok, Instagram, X and Facebook videos as timestamped JSON. When a video has no caption track, the audio is transcribed automatically.

- Auth: `apiKey` (bearer token)
- HTTPS: Yes
- CORS: No, by design. Keys are secret bearer tokens, so the API is meant for server-side use; browser apps should proxy through their own backend
- Free tier: 100 credits per month, no card required
- Docs: https://transcriptfetch.com/docs
- OpenAPI: https://transcriptfetch.com/api/v1/openapi.json

One entry, placed alphabetically between Trakt and TVDB in the Video section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2048,6 +2048,7 @@ API | Description | Auth | HTTPS | CORS |
 | [TMDb](https://www.themoviedb.org/documentation/api) | Community-based movie data | `apiKey` | Yes | Unknown |
 | [TrailerAddict](https://www.traileraddict.com/trailerapi) | Easily embed trailers from TrailerAddict | `apiKey` | No | Unknown |    
 | [Trakt](https://trakt.docs.apiary.io/) | Movie and TV Data | `apiKey` | Yes | Yes |
+| [TranscriptFetch](https://transcriptfetch.com/docs) | Timestamped transcripts for YouTube, TikTok, Instagram, X and Facebook video, server-side only | `apiKey` | Yes | No |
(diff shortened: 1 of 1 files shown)


daltino (commented):
**DECISION: COMMENT**

The addition of the TranscriptFetch API seems useful, particularly for retrieving video transcripts across various platforms, but a few details will need clarification. Can you confirm whether the API offers a free tier or full free access, as required by the contribution guidelines? Also, it might help to clarify the "No" for CORS—this could limit certain use cases if the API is not accessible from client-side applications.

Transcript-Fetch-Admin:
Thanks for taking a look.

Free tier: yes, 100 credits per month, no card required. One credit is one successful response, and failed, blocked or empty results are never charged.

On CORS: the No is intentional. Requests authenticate with a secret bearer key tied to a billable account, so the API is meant to be called server-side; sending CORS headers would invite putting the key in front-end code where any visitor could copy it. Browser apps should proxy through their own backend, the same pattern as other secret-key APIs. I have updated the entry description to say server-side only so the table communicates this at a glance.

Transcript-Fetch-Admin:
I have updated the entry wording to address both points, and the description now says server-side only so the CORS posture is clear from the table itself. Let me know if you would like anything changed.

PR #6725, opened 2026-07-30
Add Quant Data to Finance

celineycn:
| Field | Value |
|---|---|
| API | [Quant Data](https://quantdata.uk/docs) |
| Auth | No — every endpoint answers 10 calls/day per source with no key or signup (an optional key lifts the limit) |
| HTTPS | Yes |
| CORS | No |

Measured market statistics (day-type probabilities, volume-wave events with measured win rates, options max pain, dealer gamma). Genuinely usable for free without any credential: `curl https://api.quantdata.uk/v1/maxpain/SPY` returns the full JSON. Alphabetical position in the Finance table (promoted first row skipped).


daltino (commented):
This is a solid addition, and the formatting aligns with guidelines. However, since "Quant Data" includes metrics like win rates and dealer gamma, it's worth confirming whether these features have a free access tier as required. If there’s full free access or a sufficient free tier, the addition seems fitting!

PR #6764, opened 2026-08-04
Correct Food Info entry description

Some-of-the-things:
Corrects the description on the Food Info entry added in #6758.

It said "six national food composition datasets". The correct figure is five (USDA FoodData Central, McCance & Widdowson's CoFID, ANSES Ciqual, DTU Frida and AUSNUT), plus Open Food Facts, which is crowdsourced rather than national. The published source list is at https://food-info.org/data-sources

Also drops "millions of foods", which described the full catalogue including branded products rather than the reference foods the description is about.


Some-of-the-things:
The failing Validate Markdown format check looks pre-existing rather than something in this diff.

It reports 557 errors starting at line 11, and none on line 902 where this change is. The same job is also failing on pushes to master from before this PR was opened. Most of the errors read like `:--- is not a valid HTTPS option`, which suggests the validator is reading table separator rows as data rows rather than anything being wrong with a particular entry.

This is a one line description correction to an entry merged in #6758, so the diff is a single line. Happy to rebase or reword if you would like anything changed.

PR #6800, opened 2026-08-08
Add AmberOne API

scubamike124:
Adds AmberOne to **Development**.

Turns a URL into a build-ready Android, iOS, PWA, Electron or Capacitor project.

- **Docs:** https://scubamike124.github.io/amberone-api/
- **OpenAPI 3.1:** https://hq.amberoneai.com/api/v1/openapi.json (public, no key needed)
- **Source & SDKs:** https://github.com/scubamike124/amberone-api (MIT)

**Free tier:** yes — 3 wraps/month, and keys beginning `wrap_test_` run the
full pipeline without being billed, so the API can be evaluated end to end
for free.

**Auth:** `apiKey` (Bearer token).
**HTTPS:** yes.
**CORS:** no — verified. The spec endpoint sends `Access-Control-Allow-Origin: *`,
but the data endpoints deliberately do not, because calling them from a browser
would expose the caller's key.

Alphabetical placement and the 100-character description limit both checked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -533,6 +533,7 @@ API | Description | Auth | HTTPS | CORS |
 | [24 Pull Requests](https://24pullrequests.com/api) | Project to promote open source collaboration during December | No | Yes | Yes |
 | [Agify.io](https://agify.io) | Estimates the age from a first name | No | Yes | Yes |
 | [Amazonscraperapi](https://amazonscraperapi.com) | Amazon product, search & batch scraping API with residential proxies (1000 free) | `apiKey` | Yes | No |
+| [AmberOne](https://scubamike124.github.io/amberone-api/) | Turn any website into a build-ready Android, iOS, PWA or Electron app project | `apiKey` | Yes | No |
 | [API Grátis](https://apigratis.com.br/) | Multiples services and public APIs | No | Yes | Unknown |
 | [ApicAgent](https://www.apicagent.com) | Extract device details from user-agent string | No | Yes | Yes |
 | [ApiFlash](https://apiflash.com/) | Chrome based screenshot API for developers | `apiKey` | Yes | Unknown |


scubamike124:
Withdrawing this myself — it does not meet the contribution policy, and I should
have checked before opening it.

CONTRIBUTING asks that an API have "full, free access or at least a free tier."
This one does not. Entry pricing is $79/month, and the only $0 plan in our
catalogue is an internal load-test plan, not something a customer can buy.

There is a free evaluation path — signup is free, free accounts can issue keys,
and sandbox keys are never billed — but those keys run against fixtures rather
than producing real output, so calling it a free tier would be stretching the
word past what the policy means by it.

Closing rather than leaving it in the queue, since reviewer time here is scarce
and this would only have ended the same way. Apologies for the noise.

If a genuine free tier ships later, I will open a fresh PR and say so plainly.

scubamike124:
Reopening — I closed this on a wrong reading of our own product, and the policy
is in fact met.

I closed it saying there was no free tier because our plan table has no $0 row.
That was the wrong place to look. The free tier is not a purchasable plan, it is
what any account without a subscription resolves to, and it grants live output:

- Signup is free and open, no card.
- A free account gets **3 live builds per month**, PWA output.
- The build is real. Not a sandbox, not fixtures, not a watermarked preview.

Rather than assert that again, I ran it end to end against production just now,
from a fresh account with no subscription:

```
plan=free  wrapsPerPeriod=3  platforms=PWA
job 1: HTTP 202    job 2: HTTP 202    job 3: HTTP 202
job 4: HTTP 422  quota_exceeded
GET /download -> 200  application/zip  6239 bytes
  pwa/manifest.webmanifest   pwa/index.html   pwa/sw.js
  pwa/icons/icon-192.png     pwa/icons/icon-512.png
  README.md   scan-report.json   certification.json
```

An installable PWA with a manifest, a service worker and icons — the same output
a paid account receives, capped at three a month rather than limited in kind. The
fourth is refused, so the number in the listing is enforced rather than
aspirational.

Paid plans buy volume and the other targets (Capacitor, Electron, Android and iOS
project output). The free tier is not a demo of the API; it is the API.

The entry itself is unchanged from the original submission, and `CORS: No` is
still accurate — I re-checked the response headers rather than assume.

Apologies for the churn. Closing something on an unverified claim and reopening
it is worse than getting it right once, and both directions here were mine.

ArindamOSS:
Thanks for the clarification

PR #6835, opened 2026-08-12
Fix OrbitalWiki entry: Auth is No, not apiKey

Anomo1:
Correction to an existing entry in Science & Math, added in #6651. I maintain the API and got the `Auth` column wrong in that PR.

**The correction**

```diff
-| [OrbitalWiki](https://orbitalwiki.com/developers) | Catalog of 16,000+ satellites merging CelesTrak, GCAT, Wikidata; free tier included | `apiKey` | Yes | Yes |
+| [OrbitalWiki](https://www.orbitalwiki.com/developers) | Catalog of 16,000+ satellites with the source cited for every field: orbital elements, operators, NORAD/COSPAR IDs | No | Yes | Yes |
```

**Why `Auth: No` is the right value**

The public endpoints take no credentials at all:

```
$ curl -s -o /dev/null -w '%{http_code}' 'https://www.orbitalwiki.com/api/v1/satellites?limit=1'
200
```

A key exists but only raises the rate limit, which matches how NASA and other entries in this section are listed. Listing it as `apiKey` tells readers they need to sign up before they can call it, and they do not.

**Two smaller fixes in the same line**

- URL now points at `www.`, which is the canonical host. The old one served a 308 redirect on every request.
- Description says what the entry actually offers instead of "free tier included", which read as marketing. Every field in a response names the source it came from (CelesTrak, GCAT, Wikidata or SatNOGS) and returns `unknown` where the sources are silent, rather than filling the gap with a guess. That is the part a reader scanning this table would want to know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1642,7 +1642,7 @@ API | Description | Auth | HTTPS | CORS |
 | [Open Notify](http://open-notify.org/Open-Notify-API/) | ISS astronauts, current location, etc | No | No | No |
 | [Open Science Framework](https://developer.osf.io) | Repository and archive for study designs, research materials, data, manuscripts, etc | No | Yes | Unknown |
 | [OpenAlex](https://docs.openalex.org/) | Open catalog of scholarly works, authors, institutions, sources, and concepts | No | Yes | Yes |
-| [OrbitalWiki](https://orbitalwiki.com/developers) | Catalog of 16,000+ satellites merging CelesTrak, GCAT, Wikidata; free tier included | `apiKey` | Yes | Yes |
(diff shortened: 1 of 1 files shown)


Anomo1:
Note on the red check: `Validate README.md changes` is failing on `master` itself, not on this change.

The validator flags lines 2058 and 2110, the separator rows under `### Video` and `### Weather`. Both header rows are missing their leading pipe:

```
API | Description | Auth | HTTPS | CORS |
|:---|:---|:---|:---|:---|
```

so the validator reads `:---` as an Auth value. The same failure appears on the merge commits for #6776 and #6783.

This PR touches one line, 1645, in Science & Math. Happy to add the two missing pipes here if that is wanted, or to leave it as a one-line correction and let the format fix land separately.

PR #6876, opened 2026-08-15
Add Seismic Risk by Address API

ThryveBaseline:
Adds the Seismic Risk by Address API to the Science & Math section.

**What it does.** Returns USGS earthquake hazard values for a US address or coordinate: the ASCE 7 Seismic Design Category and the design ground motions. Answers "how much seismic hazard is at this location" rather than "what earthquakes happened", which is what the existing USGS entries cover.

**Free, keyless, CORS open.** No account, no API key. `Access-Control-Allow-Origin: *`, so it works client-side.

```
curl "https://seismic-risk-usgs.netlify.app/seismic-risk?lat=37.7749&lon=-122.4194"
```

Docs: https://seismic-risk-usgs.netlify.app/api-docs.html
OpenAPI: https://seismic-risk-usgs.netlify.app/openapi.json

Auth `No`, HTTPS `Yes`, CORS `Yes`, all verified against the live endpoint today. Entry is placed alphabetically between Remote Calc and SHARE. Built on the USGS Seismic Design Web Services, which already has sibling entries in this section.


xiaolaobantech:

PR #6878, opened 2026-08-15
Add Typical Weather API

ThryveBaseline:
Adds the Typical Weather API to the Weather section.

**What it does.** Returns typical monthly weather for any US ZIP from the NOAA 1991-2020 US Climate Normals: average temperature and precipitation by month. Answers "what is the weather usually like here in March" rather than "what is the forecast", which the existing Weather entries already cover.

**Free, keyless, CORS open.** No account, no API key. `Access-Control-Allow-Origin: *`, so it works client-side.

```
curl "https://typical-weather-api.netlify.app/typical-weather?zip=80202"
```

Docs: https://typical-weather-api.netlify.app/api-docs.html
OpenAPI: https://typical-weather-api.netlify.app/openapi.json

Auth `No`, HTTPS `Yes`, CORS `Yes`, all verified against the live endpoint today. Entry is placed alphabetically between Tomorrow and US Weather.


xiaolaobantech:

PR #6882, opened 2026-08-15
Add CSR.plus API

xusteve:
Adds [CSR.plus](https://csr.plus/docs/api) to the Security section.

- Free, no authentication required
- Generate certificate signing requests (CSRs) and private keys as JSON in a single POST request
- HTTPS and CORS enabled
- Open source (MIT), docs at https://csr.plus/docs/api


xiaolaobantech:

PR #6963, opened 2026-08-19
Add CompoundPulse Proof API (Finance)

22emmanueljean-glitch:
Adds a keyless finance endpoint to the Finance section.

Returns a dated verdict, the price level that invalidates the idea, and every scored factor behind it, for US stocks, major ETFs and major crypto pairs.

- **No auth, no signup.** HTTPS, CORS `*`
- Example: `curl https://www.compoundpulse.io/api/proof/NVDA`
- The publisher also publishes its own calibration — how often the score is wrong — at https://www.compoundpulse.io/research, and a full trade record including losses at https://www.compoundpulse.io/track

Checked against CONTRIBUTING.md before opening:
- Alphabetical placement (between Citi and CongressInvests)
- 6-column format matching neighbouring rows
- Description 85 chars (under the 100 limit)
- `scripts/validate/format.py` reports **557 errors before and after** this change — zero introduced
- Link returns 200


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #6980, opened 2026-08-20
Add titelia API

florianmousseau:
Adds `titelia` to **Finance**.

It answers which funds hold a listed company, and what any fund holds - read from the filings funds make to their regulators (SEC, CVM, CNMV, Finansinspektionen, SEBI) and from the annual reports European funds publish themselves, which is the half that is not otherwise republished.

- Docs: https://titelia.com/en/api
- Contract: https://titelia.com/api/openapi.json (OpenAPI 3.1)
- Auth: none, no account, no quota to apply for. HTTPS and CORS: yes.
- Free, in full, with no paid tier to upsell.

Example: https://titelia.com/api/company/FR0000120073?limit=5


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #6984, opened 2026-08-21
Add GreenCalculus API

jeremiahsay:
Adds GreenCalculus to **Environment**.

Greenhouse-gas emission factors and carbon footprint calculations over REST. Every value is returned with the source it came from and the version of the dataset, so a result can be cited and re-run to the same number later.

| Field | Value | Basis |
|---|---|---|
| Auth | `apiKey` | Bearer key; free tier, no card required, and several endpoints need no key at all |
| HTTPS | Yes | HTTPS only |
| CORS | Yes | verified — `access-control-allow-origin: *` on `api.greencalculus.com/v1/factors` |

Docs: https://greencalculus.com/developers

Checklist against CONTRIBUTING:

- Placed alphabetically in Environment, between *Danish data service Energi* and *GrünstromIndex*
- Description is 94 characters (limit 100)
- Name carries no TLD and does not end in "API"
- Columns padded with one space either side
- One link, one commit, targeting `master`
- Searched existing issues and PRs first — no prior GreenCalculus entry


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7004, opened 2026-08-22
Update Orizn Visa entry: stale counts, link to docs

MattJeff:
Correction to an existing entry under **Transportation**.

The description read *"199 countries, 39K+ passport-destination pairs"*. That count was frozen from an early release and no longer matches the dataset — the project has since removed hardcoded counts from its published metadata rather than reprint a number that goes stale again. Live coverage is public and unauthenticated at `https://visa.orizn.app/api/v1/visa/stats`.

The link now points at the documentation instead of the marketing page.

`Auth`, `HTTPS` and `CORS` values are unchanged and still accurate: an API key is required (free tier, no card), HTTPS only, and `Access-Control-Allow-Origin: *` on every public endpoint.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7010, opened 2026-08-22
Fix false positives in format and link validators

aljihadjundy-oss:
## What this fixes

Both validators report large numbers of failures that are artifacts of their own implementation rather than problems with `README.md`.

### `format.py`: 557 errors on an unmodified README

Two causes:

1. The separator-row guard tests for `|---`, but every table in the README writes its separator as `|:---|`. All 45 separator rows are therefore validated as API entries, each producing five bogus column errors.
2. The scan covers the whole file, so the three-column sponsor table and the promo headings above `## Index` are checked against the five-column entry schema.

This PR replaces the prefix test with `is_table_row()`, which rejects any Markdown separator form, and adds `get_api_list_bounds()` to confine both passes to the `## Index` .. `## License` range. The bounds are applied as a line filter rather than a slice, so error messages keep reporting real line numbers.

The three module-level regexes are also made raw strings. They currently emit `SyntaxWarning` on Python 3.12+ and will eventually stop compiling.

**Result: 557 errors to 47.** The remaining 47 are genuine README violations, listed below.

### `links.py`: 402 links reported broken, 104 of them incorrectly

`check_if_link_is_working()` sets a `host` header derived from the original URL. `requests` already derives `Host` from the URL on every hop, so the hand-set value only takes effect where it is wrong: on a cross-domain redirect it pins the original host onto the new target, which then rejects the request with `421` or redirects again until `TooManyRedirects`.

Removing the header drops a full run from 402 reported failures to 298, eliminating all 62 `TooManyRedirects` and all 5 misdirected-request failures. Links currently reported as dead include `docs.microsoft.com`, `plaid.com/docs`, `code.gov`, `ups.com` and `solana.com`.

`get_host_from_link()` is left in place; it has direct unit test coverage.

## This does not make CI green

To be clear about scope: `format.py` still exits non-zero after this change. It goes from 557 errors to 47, and those 47 are real:

- 22 categories are not in alphabetical order
- 10 descriptions exceed the 100-character limit
- 4 bare `apiKey` auth values are missing backticks
- 2 descriptions end with punctuation, 1 is not capitalised
- 1 table header row carries a leading pipe the other 44 categories do not
- `links.py --only_duplicate_links_checker` reports 2 duplicated URLs

Those are README content fixes and belong in a separate PR. I have them prepared if they would be useful, but I did not want to mix a content change into a tooling fix.

## Verification

- `python -m unittest discover tests/` — 31 tests, all passing, no test changes required
- `python scripts/validate/format.py README.md` — 557 errors to 47
- `python scripts/validate/links.py README.md` — 402 reported failures to 298

The link numbers come from full runs over all 1693 links. Every failure classification was confirmed across three separate runs to rule out rate limiting and timeout jitter.

## Note on commit count

The contributing guide asks for a single squashed commit. I have kept these as two commits because they are independent fixes to two different files, and each is easier to review and to revert on its own. Happy to squash if you would prefer.

### DIFF
--- a/scripts/validate/format.py
+++ b/scripts/validate/format.py
@@ -24,14 +24,46 @@ num_segments = 5
 min_entries_per_category = 3
 max_description_length = 100
 
-anchor_re = re.compile(anchor + '\s(.+)')
-category_title_in_index_re = re.compile('\*\s\[(.*)\]')
-link_re = re.compile('\[(.+)\]\((http.*)\)')
+anchor_re = re.compile(anchor + r'\s(.+)')
+category_title_in_index_re = re.compile(r'\*\s\[(.*)\]')
+link_re = re.compile(r'\[(.+)\]\((http.*)\)')
+table_separator_re = re.compile(r'^\|[\s:|-]+$')
+
+# The API listing lives between these two headings. Everything outside of it
+# (sponsor banners, promo tables, the license footer) is not an API entry and
+# must not be validated as one.
+list_start_heading = '## Index'
+list_end_heading = '## License'
 
 # Type aliases
 APIList = List[str]
 Categories = Dict[str, APIList]
 CategoriesLineNumber = Dict[str, int]
+
+
+def get_api_list_bounds(lines: List[str]) -> Tuple[int, int]:
+    """Return the [start, end) line range holding the API listing."""
+
+    start = 0
+    end = len(lines)
+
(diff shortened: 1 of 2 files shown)


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7020, opened 2026-08-23
Add Offendersearch to Government

Clavaa:
Adds Offendersearch under Government (US public-records data), placed alphabetically between Neotimo and Open Government. It searches all 58 US sex-offender registries in one API call and returns scored records with a link to the official registry entry behind each match. Free tier (25 searches, no card), apiKey auth, HTTPS. Docs: https://offendersearch.app/docs


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7022, opened 2026-08-23
Add StackScan API

Awebus:
Adds StackScan to **Security**, alphabetically between SSL Domain Health Check and Threat Jammer.

**Disclosure: I work for StackScan.**

The API answers two directions over an index of 360M+ sites and 55,000+ technologies:

- `domains/lookup` returns the technology stack for a domain, with each technology's category and how many sites run it.
- `technologies/lookup` goes the other way, returning site counts and country distribution for a given technology, optionally with the matching site list.

There are also batch variants of both, and an endpoint for building and downloading full technology lists.

Verified against tesla.com: 33 technologies returned, each with parent and child category.

Free tier on email verification: 100 lifetime API credits. Paid plans add bulk volume.

- Auth: bearer token plus a workspace header, so `apiKey`
- HTTPS: yes
- CORS: yes, verified `access-control-allow-origin: *`

Description is 85 characters, under the 100 limit. `scripts/validate/format.py` reports no new errors.


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7060, opened 2026-08-25
Add RevAddress API

jtalk22:
Adds RevAddress to Data Validation.

- Docs: https://revaddress.com/docs/ — OpenAPI at https://api.revaddress.com/openapi.yaml
- Free tier: 1,000 credits a month, no card, does not expire (https://revaddress.com/pricing/)
- What it does: US address standardization (Census-based), congressional and state legislative district lookup including the 2026 redistricted lines, tract-level Census data, all with per-block provenance
- Auth: `apiKey` (X-API-Key header) · HTTPS: yes · CORS: allowlisted origins only, so "No"

Alphabetical within the section (after PurgoMalum), description under 100 characters, one link.


ArindamOSS (changes requested):
Please remove the trailing |  which is creating an empty 6th column

PR #7065, opened 2026-08-26
Add Banking Access Index to Finance

jianjettfu-oss:
Adds the Banking Access Index to the Finance section.

**What it is:** a free, no-auth JSON endpoint recording which US business banking providers accept founders by country of residence, with a sourced evidence quote behind each claim. Published under CC-BY and archived with a DataCite DOI (10.5281/zenodo.21336392).

**Endpoint:** https://www.globalsolo.global/data/banking-access-index.json

Field values were measured, not assumed:

```
$ curl -sI -H 'Origin: https://example.com' https://www.globalsolo.global/data/banking-access-index.json
HTTP/2 200
content-type: application/json; charset=utf-8
```

- Auth: No
- HTTPS: Yes
- CORS: Yes

**Checks run:**
- `scripts/validate/format.py` — the new row produces zero errors. Message-level diff against the base README is identical, so nothing new was introduced (the repo has pre-existing ordering warnings unrelated to this change).
- Alphabetical placement: after `Bank Data`, before `Billplz`.
- No existing entry covers account-opening eligibility by country of residence — the nearest neighbours (Plaid, Nordigen, Mono) all cover transaction data on accounts that already exist.

Free and non-commercial: the dataset is CC-BY, downloadable in full, and requires no signup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -859,7 +859,7 @@ API | Description | Auth | HTTPS | CORS |
 | [Alpha Vantage](https://www.alphavantage.co/) | Realtime and historical stock data | `apiKey` | Yes | Unknown | |
 | [Banco do Brasil](https://developers.bb.com.br/home) | All Banco do Brasil financial transaction APIs | `OAuth` | Yes | Yes | |
 | [Bank Data](https://apilayer.com/marketplace/bank_data-api) | Instant IBAN and SWIFT number validation across the globe | `apiKey` | Yes | Unknown | |
-| [Banking Access Index](https://www.globalsolo.global/data/banking-access-index) | Which US business banking providers accept founders by country of residence, with sourced evidence | No | Yes | Yes | |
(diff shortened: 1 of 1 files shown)


ArindamOSS (changes requested):
Please remove the trailing | | and trim the description to stay under 100 characters

jianjettfu-oss:
Thanks @ArindamOSS — both addressed in 8cf05a1:

- removed the trailing empty column, so the row now ends at the CORS field
- description trimmed to 89 characters (was 98)

`python3 scripts/validate/format.py README.md` reports no errors on this line, and its output is byte-identical to the pre-change baseline.

jianjettfu-oss:
@ArindamOSS — both points were addressed in 8cf05a1 on Aug 29 (trailing column removed, description trimmed to 89 chars, format.py clean). Could you re-review when you get a chance? No rush.

PR #7081, opened 2026-08-26
Add Window Sticker API

ryanfrigo:
Adds one entry to the Vehicle section.

**Window Sticker** — https://windowsticker.org/api-docs

Returns the original factory window sticker (Monroney label) as a PDF for a
17-character US VIN, plus a full NHTSA vPIC specification decode.

- `Auth`: `No` — no key, no registration
- `HTTPS`: `Yes`
- `CORS`: `Yes` — responses send `Access-Control-Allow-Origin: *`

Endpoints:

- `GET /api/v1/vin/{VIN}` — JSON decode and sticker availability
- `GET /api/sticker/{VIN}` — the PDF itself
- OpenAPI 3.1 description at `/openapi.json`

Example:

```
curl https://windowsticker.org/api/v1/vin/JF1GUAFC1T8203971?sticker=false
```

Original PDFs are available for 16 makes — the GM, Ford and Stellantis brands plus
Subaru, Kia, Hyundai and Genesis. Every other VIN returns the specification decode
only, which the docs state plainly.

Checklist:

- [x] One link added
- [x] Alphabetical order within the section (after Smartcar)
- [x] Description is 63 characters
- [x] Single squashed commit


ryanfrigo:
Rebased onto current master — the Vehicle table gained TransparentCars and Wheelwise since this was opened, so the row now sits after Wheelwise and the section is back in alphabetical order. Single commit, one line added.

PR #7086, opened 2026-08-27
Add Urgences Québec (Health)

Arephan:
Adds one row to Health.

Urgences Québec is an independent open API over Quebec's emergency-room situation. Québec's health ministry publishes a single file that it overwrites every hour, so nothing historical survives on the official side; this API keeps the hourly readings and serves them as JSON.

- `/api/now` — current occupancy, stretcher counts, people waiting, patients past 24h/48h and average stay, for all 120 emergency rooms
- `/api/facilities` — addresses, coordinates, phone, region for the same 120
- `/api/cliniques`, `/api/chirurgies` — walk-in clinic hours and surgical wait times

No key, no rate limit, `Access-Control-Allow-Origin: *`, HTTPS only. Data page with the sources and licence: https://sante.handled.tools/en/donnees

Not affiliated with the ministry — the underlying data is its open publication.


zhangmengmeng0777:
stable and low‑cost tokens：xiaoyouling.top

PR #7092, opened 2026-08-27
Add Shotium API

shery:
Adds Shotium to the Development section, alphabetically between Sheetsu and SHOUTCLOUD.

Shotium renders any public URL to a screenshot, or a 1200x630 Open Graph image from a built-in template.

Column values, for the record:

- **Auth** `apiKey` — a Bearer API key.
- **HTTPS** Yes — HTTP is not served.
- **CORS** No — verified against production rather than assumed. The key belongs on the server side, so no `Access-Control-Allow-Origin` headers are sent.

Documentation: https://shotium.com/docs
OpenAPI 3.1 definition: https://shotium.com/openapi.json

Ran `scripts/validate/format.py` and `scripts/validate/links.py --only_duplicate_links_checker` locally: the new entry raises no errors, and the error count is unchanged from master (the remaining ones are pre-existing).


zhangmengmeng0777:
stable and low‑cost tokens：xiaoyouling.top

l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #6166, opened 2026-05-27
Add Audexum TTS API to Text Analysis section

audexum:
Added Audexum to the Text Analysis section.

- API: https://audexum.com/docs
- Auth: apiKey (Bearer token)
- HTTPS: Yes
- CORS: Yes
- Description: Text-to-speech REST API, 43 voices, 33 languages, freemium

PR #7091, opened 2026-08-27
docs: fix hyphenation 'open-source icon sets'

deepshekhardas:
Fixes hyphenation for Iconify entry: 'open source icon sets' -> 'open-source icon sets'.

PR #7096, opened 2026-08-27
Correct the Audexum entry and move it to Machine Learning

audexum:
Correcting our own entry (added in #6166).

**Section.** It is currently under `Text Analysis`, between `apilayer languagelayer` and `Cloudmersive Natural Language Processing` — language-detection and NLP text tools. Audexum is a speech API, so `Machine Learning` is the closer fit under the "section most in line with the services offered" rule; that section already lists BRAINIALL (audio transcription) and Hugging Face's audio inference. It was also out of alphabetical order where it sat (`Audexum` after `Aylien`); the new position, between `AI For Thai` and `BRAINIALL`, is correct.

**Description.** The old text described only text-to-speech and did not mention speech-to-text at all, which is half of what the API does. It also said 33 languages where the docs say 32.

```
- | [Audexum](https://audexum.com/docs) | Text-to-speech REST API with 43 voices and 33 languages | `apiKey` | Yes | Yes |
+ | [Audexum](https://audexum.com/docs) | Speech-to-text in 25 languages and text-to-speech with 43 voices in 32 languages | `apiKey` | Yes | Yes |
```

Description is 80 characters. Auth, HTTPS and CORS are unchanged and still accurate (`access-control-allow-origin: *`). Free tier is unchanged and requires no purchase. One line changed, no new `scripts/validate/format.py` findings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1355,6 +1355,7 @@ API | Description | Auth | HTTPS | CORS |
 |:---|:---|:---|:---|:---|
 | [AI Economics Tools](https://piszczek.pl/tools/api) | Token cost, LLM energy, agent-hour and Proof-Adjusted Autonomy calculators by Michał Piszczek | No | Yes | Yes |
 | [AI For Thai](https://aiforthai.in.th/index.php) | Free Various Thai AI API | `apiKey` | Yes | Yes |
+| [Audexum](https://audexum.com/docs) | Speech-to-text in 25 languages and text-to-speech with 43 voices in 32 languages | `apiKey` | Yes | Yes |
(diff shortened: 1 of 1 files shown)


audexum:
Note on the red check, in case it reads as a broken submission: the `Validate README.md changes` failures are pre-existing and not from this line.

Every annotation lands on `:---` table-separator rows — L2081, L2098, L2153 — and none on L1358 or L1930, which are the only two lines this PR touches. Running `scripts/validate/format.py README.md` locally gives an **identical** set of findings before and after the change (564 either way).

PRs #7090 and #7091, both one-word doc fixes, show the same `FAILURE, SUCCESS` pattern right now, so the validator appears to be failing against `master` for everyone rather than against any particular contribution.

Happy to rebase once that is sorted, or to close this if you would rather take the description fix and the section move as two separate PRs.

PR #7178, opened 2026-09-01
Add receivemail.dev to Email

flovoice53-tech:
Adds `receivemail.dev` to the **Email** section.

- Disposable/ephemeral inbox API, in the same family as DropMail, Guerrilla Mail, mail.tm and mail.gw which are already listed.
- Fully free, no account. `POST /mailboxes` needs no authentication and returns an address plus a per-mailbox token used to read that mailbox's messages.
- HTTPS only. No CORS headers (marked `No`).
- Not device- or purchase-gated.

Format matches the existing 5-column table; entry placed alphabetically between PostStack and Sendgrid.


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7184, opened 2026-09-01
fix(ci & validate): RFC 3986 urlparse host extraction, bash diff parser bug, and format fixes

subbareddypalagiri:
### Summary of Changes

This Pull Request provides high-impact fixes and hardening for the validation and CI suite:

1. **RFC 3986 URL Host Extraction (`links.py`)**:
   - Replaced fragile custom string slicing with standard `urllib.parse.urlparse`.
   - Properly handles URLs containing authentication credentials (`user:pass@`), custom ports (`:8080`), IPv6 addresses, queries, and fragments without invalid Host headers.
   - Added unit test cases in `test_validate_links.py`.

2. **Git Diff Additions Parsing Bug (`github_pull_request.sh`)**:
   - `cat diff.txt | egrep "+"` was matching non-added lines (such as diff headers `+++ b/README.md`, C++ references, or math symbols).
   - Fixed to `grep '^+[^+]' diff.txt | sed 's/^+//'` to isolate genuine added lines and prevent false link validation failures in CI.

3. **Bash `set -e` Error Handling Fix (`github_pull_request.sh`)**:
   - Because `set -e` is active, a failure in `python scripts/validate/links.py` exited immediately before reaching `if [[ $? != 0 ]]`.
   - Restructured to idiomatic `if ! python scripts/validate/links.py "$LINK_FILE"; then ...`.

4. **Punctuation Parentheses Replacement Fix (`format.py`)**:
   - Fixed `punctuation.replace('()', '')` no-op bug by replacing with `punctuation.replace('(', '').replace(')', '')`.
   - Added raw string regex prefixes (`r'...'`) to eliminate Python 3.12+ `SyntaxWarning` messages.

### Verification

All 31 unit tests pass locally with 0 warnings:
```bash
$env:PYTHONPATH="scripts"; python -m unittest discover -s scripts/tests -t scripts
Ran 31 tests in 0.003s — OK
```

### DIFF
--- a/scripts/github_pull_request.sh
+++ b/scripts/github_pull_request.sh
@@ -37,7 +37,7 @@ curl -L "$DIFF_URL" -o diff.txt
 echo "------- BEGIN DIFF -------"
 cat diff.txt
 echo "-------- END DIFF --------"
-cat diff.txt | egrep "\+" > additions.txt
+grep '^+[^+]' diff.txt | sed 's/^+//' > additions.txt
 
 echo "------ BEGIN ADDITIONS -----"
 cat additions.txt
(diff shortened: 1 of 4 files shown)


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7212, opened 2026-09-03
Add PVGIS API

deepshekhardas:
Add [PVGIS](https://joint-research-centre.ec.europa.eu/pvgis-photovoltaic-geographical-information-system/getting-started-pvgis/api-non-interactive-service) to Environment, alphabetically between PM25.in and PVWatts.

- Free, no auth (per-IP rate limit only)
- HTTPS: Yes, CORS: No (docs state AJAX access is not allowed)
- Complements the listed PVWatts entry, which requires an API key


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7214, opened 2026-09-03
Add Semantic Scholar API

deepshekhardas:
Add [Semantic Scholar](https://api.semanticscholar.org/api-docs/) to Science & Math, alphabetically between Remote Calc and SHARE.

- Free, no auth (most endpoints public without authentication)
- HTTPS: Yes, CORS: Yes
- Complements the listed arXiv, OpenAlex and CORE entries


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7241, opened 2026-09-04
Remove Chuck Norris Database API (domain repurposed)

AdmiraLahav:
removed:
| [Chuck Norris Database](http://www.icndb.com/api/) | Jokes | No | No | Unknown |

as the site is no longer a Chuck Norris jokes site but a gambling site


AdmiraLahav:
created another pull request with another issue

PR #7248, opened 2026-09-05
Removed repurposed site domains

AdmiraLahav:
removed a changed domain:
http://www.icndb.com/api/

as this once was a Chuck Norris jokes site but now is a gambling site

also, the domain:
https://catboys.com/api

is for sale so the site does not exist anymore


AdmiraLahav (commented):
no conflicts created

l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7269, opened 2026-09-06
Add Malwagon API

projectboot:
Adds the Malwagon API to the Anti-Malware section.

It submits a file or URL for detonation on instrumented Windows and Linux virtual machines and returns the behaviour report, extracted indicators and ATT&CK technique mappings. Documentation is at https://malwagon.com/docs/api

There is a free community tier: an API key is issued without payment details, which is why the entry is listed as `apiKey` rather than requiring a purchase. Placed alphabetically between MalShare and MalwareBazaar, one entry, targeting master.


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7275, opened 2026-09-08
feat: add DataCite API

deepshekhardas:
feat: add DataCite API

Search and retrieve DOI metadata for research datasets and publications (No/Yes/Yes) in Science & Math, alphabetically between CycleCalcs and Europe PMC.


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7281, opened 2026-09-08
Add Motomarks to Vehicle

aciaschetti:
Vehicle already covers specs, prices, and telematics. This row is the manufacturer logo image CDN those listings sit next to.

Docs: https://motomarks.io/docs
Auth apiKey. CORS Unknown. HTTPS Yes.
Free tier exists (1,000 requests/day).


aciaschetti:
Closing this. Motomarks is already submitted on the community list instead, so this duplicate should not stay open.

PR #7357, opened 2026-09-13
Remove gridcarbon API

gupeng1208:
The gridcarbon API was shut down on 2026-09-13. Every endpoint now returns HTTP 410, so this entry no longer points to a working free API.

I added the entry in #7224; this removes it again. One line, one commit.


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7358, opened 2026-09-13
Add Kavel API

hanshs474:
Adds Kavel to **Photography**: text-to-image and image editing over HTTPS with no authentication.

- Auth: `No` — callers send any string as `x-anon-id`
- HTTPS: `Yes`
- CORS: `No` (no `Access-Control-Allow-Origin` on responses)

Docs with raw HTTP examples: https://kavel.readthedocs.io/en/latest/sdk/http/


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7359, opened 2026-09-14
fix(validate): skip |:--- table separator rows in format check

Autumn-Atlas:
## Problem

The "Validate Markdown format" CI check currently fails with ~500 phantom errors that point at table separator rows, e.g.:

```
(L129) description should not end with -
(L129) each segment must start and end with exactly 1 space
(L129) :--- is not a valid Auth option
```

## Root cause

The skip condition in `scripts/validate/format.py` only ignores separator rows starting with `|---`:

```python
if not line_content.startswith('|') or line_content.startswith('|---'):
```

But 46 of the 52 tables in README.md use the left-aligned separator style `|:---|:---|…`. Those rows fall through to `check_entry()` and are parsed as API rows, producing the error storm above. (The two `startswith('|---')` checks — lines 55 and 231 — both need the extra case.)

## Fix

One-line change per site: also skip `|:---` rows.

## Verified

- Before: 566 format errors on current master
- After: 70 — the 496 phantom separator errors are gone; remaining errors are genuine content issues (40 categories out of alphabetical order, APILayer-suite section structure)


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7367, opened 2026-09-14
Add Microburbs API

David-Mib:
Adds **Microburbs** to the **Open Data** section — Australian property and suburb data (valuations, sale and rent history, comparable sales, schools, demographics, crime, risk overlays, zoning and development), keyed by address or ABS locality.

**Free access, no account required.** Send `Authorization: Bearer test` and every call is free. The sandbox covers one sample suburb and one sample property per state (Bondi, Belmont North, St Kilda (Vic.), Fortitude Valley, Subiaco, Unley, Sandy Bay, Kambah, Nightcliff); requests outside that set return 403 with a message naming the sandbox set. Being upfront: the rest of the catalogue is a paid, metered API, so the free tier is a sandbox rather than a free quota over all of Australia — but it needs no signup and is enough to build and test against:

```
curl -H 'Authorization: Bearer test' \
  'https://api.microburbs.com.au/v1/suburbs/Bondi/hero/summary'
```

Field values were checked rather than assumed:
- **Auth `apiKey`** — bearer token in the `Authorization` header.
- **HTTPS Yes** — HTTPS only.
- **CORS No** — verified; no `Access-Control-Allow-Origin` is returned for a third-party origin, so it is server-side only.

OpenAPI 3.1.0 with 173 paths: https://api.microburbs.com.au/openapi.json

- [x] My submission is formatted according to the guidelines in the [contributing guide](/CONTRIBUTING.md)
- [x] My addition is ordered alphabetically
- [x] My submission has a useful description
- [x] The description does not have more than 100 characters
- [x] The description does not end with punctuation
- [x] Each table column is padded with one space on either side
- [x] I have searched the repository for any relevant issues or pull requests
- [x] Any category I am creating has the minimum requirement of 3 items
- [x] All changes have been squashed into a single commit

PR #7375, opened 2026-09-15
Update Microburbs description

David-Mib:
Updates the description of the existing [Microburbs](https://www.microburbs.com.au/developers/api-docs) entry in **Open Data** (added in #7367). No new entry, no other rows touched.

Before:
`Australian property and suburb data: valuations, sales, rents, schools, demographics, risk, zoning`

After:
`Australian property data below the suburb: street price forecasts, crime, sales, rents and schools`

Why: the old wording led with valuations, which is the least distinctive thing the API does and is what several other providers already list. The API's actual point of difference is that it resolves below the suburb — a price forecast per street and crime for small Census areas — so the description now says that instead.

Format: 98 characters (under the 100 limit), capitalised, no trailing punctuation, columns padded with one space, alphabetical position unchanged. `scripts/validate/format.py` reports no errors on the changed line.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


David-Mib:
Note on the red `Validate README.md changes` check: it is pre-existing and not caused by this PR. The workflow runs `scripts/validate/format.py` over the whole of `README.md`, and that already fails on `master`. I ran it locally against both trees:

- `master` → exit 1, 566 errors
- this branch → exit 1, **the same 566 errors**, byte-for-byte identical

The changed line (Microburbs, Open Data) produces no errors at all: 98 characters, capitalised, no trailing punctuation, one space of padding per column, alphabetical position and link unchanged.

l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7379, opened 2026-09-15
Add OddsRadarWire API

AiEnjoyerrrr81:
Adds OddsRadarWire to Sports & Fitness.

- Docs: https://oddsradarwire.com/docs.html
- Live and prematch sports odds (tennis, soccer, cricket, basketball and other traditional sports) with the provider's raw no-vig probability on every selection.
- REST plus realtime SSE.
- Free tier: 100 REST calls/day, no card. A public demo key also exists, so the API can be evaluated with no signup at all: https://oddsradarwire.com/v1/fixtures?key=demo&limit=3
- Auth: `apiKey` (x-api-key header or key query parameter).
- HTTPS: yes. CORS: not enabled (server-side use).

The API is independent of the data provider and is not affiliated with Sportradar.


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7381, opened 2026-09-15
Add FanLine Wire to Sports & Fitness

AiEnjoyerrrr81:
Adds [FanLine Wire](https://fanlinewire.com), a real-time FanDuel Sportsbook odds feed, to Sports & Fitness.

- **Docs:** https://fanlinewire.com/docs
- **Keyless public snapshot** (no signup, one JSON document): https://fanlinewire.com/odds.json — that is why the auth column is `No`, and it is served with CORS open
- Live and in-play plus pre-match: moneyline, spread and totals as the core markets, plus the book's futures and outright markets
- Also serves a read-only MCP connector at https://fanlinewire.com/mcp and a zero-dependency Node client at https://github.com/Fanlinewire/fanlinewire-sdk

Not a bookmaker · 18+ · Not affiliated with or endorsed by FanDuel — FanDuel is a trademark of its respective owner.


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7388, opened 2026-09-15
Add CoworkingView API

erich-mueller:
Adds CoworkingView to **Open Data**.

| Field | Value |
|---|---|
| Docs | https://coworkingview.com/openapi.json (OpenAPI 3.1, 14 paths) |
| Base URL | https://api.coworkingview.com/v1 |
| Auth | `No` — no key, no signup, no rate-limited free tier to upgrade out of |
| HTTPS | Yes |
| CORS | **No** — reads are server-side only, stated honestly rather than left `Unknown` |

A read-only catalogue of coworking spaces and private offices across 63 cities in Europe and the Gulf: locations, amenities, workspace types, and the rates operators publish themselves. It also serves a per-desk price index released under CC BY 4.0, with the sample size printed beside every figure.

Checklist against CONTRIBUTING:

- Placed in `Open Data`, alphabetically between `CollegeScoreCard.ed.gov` and `CuttingToolsAI`
- Description is 90 characters
- Name carries no TLD and does not end in `API`
- One link, one file, one commit
- `scripts/validate/format.py` reports exactly the same 566 findings before and after this change; none of them is on the added line

Disclosure: I maintain this service.

PR #7389, opened 2026-09-16
Add CoworkingView API

erich-mueller:
Adds CoworkingView to **Open Data**.

Supersedes #7388, which I am closing — identical commit, moved to our organisation's fork so the personal fork can be cleaned up. Same single line, same content.

| Field | Value |
|---|---|
| Docs | https://coworkingview.com/openapi.json (OpenAPI 3.1, 14 paths) |
| Base URL | https://api.coworkingview.com/v1 |
| Auth | `No` — no key, no signup, no rate-limited free tier to upgrade out of |
| HTTPS | Yes |
| CORS | **No** — reads are server-side only, stated honestly rather than left `Unknown` |

A read-only catalogue of coworking spaces and private offices across 63 cities in Europe and the Gulf: locations, amenities, workspace types, and the rates operators publish themselves. It also serves a per-desk price index released under CC BY 4.0, with the sample size printed beside every figure.

Checklist against CONTRIBUTING:

- Placed in `Open Data`, alphabetically between `CollegeScoreCard.ed.gov` and `CuttingToolsAI`
- Description is 90 characters
- Name carries no TLD and does not end in `API`
- One link, one file, one commit
- `scripts/validate/format.py` reports exactly the same 566 findings before and after this change; none of them is on the added line

Disclosure: I maintain this service.


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7408, opened 2026-09-17
Add EU VAT Rates by Commodity Code

humora2504:
Adds one entry to Government.

**What it is.** VAT rates for all 27 EU member states, mapped to the CN (goods) and CPA (services) commodity codes each rate actually applies to, with each member state's own legal note kept verbatim. 2,667 commodity codes, 1,120 rate rows. Rebuilt daily.

Published VAT tables give a country's standard rate and its reduced rates. They do not say which goods each reduced rate covers, which is the part that decides what you charge. The European Commission publishes that mapping, but only through a SOAP service. This is the same content as plain JSON.

**Free access, no tier, no purchase.** No key, no account, no rate limit, no paid plan of any kind on the API:

```bash
curl https://humora2504.github.io/eu-vat-by-commodity-code/data/by-code/4901.json
curl https://humora2504.github.io/eu-vat-by-commodity-code/data/by-country/DE.json
curl https://humora2504.github.io/eu-vat-by-commodity-code/data/all.json
```

**Fields verified rather than assumed.** Auth `No`: no credential of any kind. HTTPS `Yes`. CORS `Yes`: confirmed with a real cross-origin `fetch` from a different domain, and the response carries `access-control-allow-origin: *`.

**Licence.** The data is the European Commission's, CC BY 4.0 under Decision 2011/833/EU, attributed in every file. The scripts that build it are MIT.

**Format checks.** Inserted alphabetically in Government, description 71 characters, auth value in backticks. I ran `scripts/validate/format.py` before and after: 567 findings before my change, 567 after, none on my line.

**Disclosure, so a maintainer does not have to discover it.** I maintain this dataset, and on the project's site I sell a change feed filtered to a customer's own commodity codes. The API being added here is not part of that: it has no paid tier, nothing is gated, and the data is free for commercial use because it is public data. If that is over the line for this list, say so and I will close this without argument.


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7409, opened 2026-09-17
Add Litport Free Proxy API

litport:
Adds the Litport free proxy API to **Development**, alphabetically between `License-API` and `Logs.to`.

| API | Description | Auth | HTTPS | CORS |
| --- | --- | --- | --- | --- |
| [Litport Free Proxy](https://litport.net/docs/free-proxy-api) | Live-tested HTTP, SOCKS4, and SOCKS5 proxies with latency, uptime, and geolocation | No | Yes | No |

**Disclosure:** I maintain this API, so I want to be upfront given the guidance about marketing PRs. The endpoint is genuinely free and fully public — no key, no account, no signup, no paid tier gating it, and no device purchase. You can verify it right now:

```sh
curl "https://litport.net/api/free-proxy/snapshot?checkedWithinMin=60"
```

Returns JSON with each proxy's protocol, host, port, country/region/city, ASN, anonymity level, HTTPS support, measured latency, and 24h/7d uptime with check counts. Refreshed every few minutes.

Field checks against CONTRIBUTING.md:
- **Auth `No`** — verified, the request above carries no credentials.
- **HTTPS `Yes`** — verified.
- **CORS `No`** — I checked rather than assumed; the response carries no `Access-Control-Allow-Origin`, so it is server-side only for now.
- Description is 82 characters, under the 100 limit; name carries no TLD and does not end in "API".
- One link, one commit, alphabetical placement, existing column format.

Comparable entries already listed in the same section include `ProxyForge` and `ProxyKingdom`. Happy to adjust the wording, section, or drop it if you would rather not carry it.


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7417, opened 2026-09-18
Add TAPAC MCP — B2B contact search with SMTP verification in the same call

axelfreeman:
## What this adds

[TAPAC](https://tapacapi.com) — an MCP server that finds B2B business contacts and verifies every email over SMTP **inside the same call**, so returned contacts bounce at 2–5% instead of the 10–35% typical of stored databases.

**Endpoint (remote, streamable HTTP):** `https://tapacapi.com/mcp`
**Local install:** `npx -y @tapacapi/mcp`
**Auth:** API key as Bearer token. Free tier: 100 searches, no card — https://tapacapi.com/get-key
**Tools:**
- `tapac_find_contacts` — industry, job_titles, company_size, location, source (`website` | `telegram` | `discord`), limit → contacts with SMTP-verified emails and the source each was found in
- `tapac_status` — server version, key state, remaining quota (returns onboarding rather than an error when no key is set)

**Docs / manifests:** https://tapacapi.com/mcp.html · https://tapacapi.com/developers · OpenAPI 3.1: https://tapacapi.com/openapi.json · repo: https://github.com/axelfreeman/tapac-mcp (MIT)

Live check: `POST https://tapacapi.com/mcp` with `initialize` returns HTTP 200 and `serverInfo.name = tapac`.

One row added to an existing table in this list, using the columns already present. Pricing is the vendor-published pay-per-use rate (100 free searches, then $0.10–0.50 per verified contact).


axelfreeman:
Closing this myself — the row landed in the wrong table on this repo (it is not an MCP/agent tool directory), so it does not belong here. It was added by an automated discovery step with too loose a filter; the filter is fixed.

PR #7418, opened 2026-09-18
Add NameGender API

anpekesen:
Adds NameGender to the Development section.

**What it does:** returns the likely gender of a first name, a full name, an email address or a username, with a probability and the sample size the probability is based on.

**Free access:** 100 credits per day on a free account, refilled daily, no card required. Full API access on the free tier, including bulk lookups.

| Field | Value |
|---|---|
| Auth | `apiKey` (Bearer or `X-Api-Key`) |
| HTTPS | Yes |
| CORS | Yes (`Access-Control-Allow-Origin: *`, verified with a preflight request) |
| Docs | https://namegender.com/docs |

Checked against CONTRIBUTING:

- Placed alphabetically between `MyIPRightNow` and `Nationalize.io` in Development
- Description is 72 characters
- One link, not previously listed, no TLD in the name
- `scripts/validate/format.py` reports the same 573 pre-existing findings before and after this change, and none on the added line


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7428, opened 2026-09-19
Add CuratorSearch API

nsamman:
Museum, gallery and curatorial vacancies gathered daily from 400+ institutions' own careers pages, with every advertised salary archived.

- Free, no authentication, no key
- HTTPS, and `access-control-allow-origin: *`
- Documented at https://curatorsearch.com/developers with an OpenAPI 3.1 description at https://curatorsearch.com/openapi.json
- Also served as RSS and as an MCP endpoint

Example: https://curatorsearch.com/api/jobs?q=curator&limit=5


l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

PR #7442, opened 2026-09-21
Add NameGender MCP server

anpekesen:
Adds the NameGender MCP server to the **MCP Servers** section.

| Field | Value |
| --- | --- |
| Repo | https://github.com/anpekesen/namegender-mcp (MIT, public) |
| Auth | `apiKey` — free tier available, 100 lookups/day, no card |
| Transport | `stdio` only — the server connects a `StdioServerTransport` and implements no HTTP or SSE transport |
| Install | [Glama](https://glama.ai/mcp/servers/anpekesen/namegender-mcp) — verified reachable; not listed on Cursor or Anthropic, so those are omitted |

It resolves a first name, full name, email address or username to a gender and returns the evidence with it: probability, sample size, country and source. It returns `unknown` when the evidence is insufficient rather than guessing.

One entry, one link. The NameGender REST API is already listed under Development (#7418); this is the separate MCP server, not a new version of that entry.


anpekesen:
Note on the red `Validate README.md changes` check — it is not caused by this diff.

`scripts/validate/format.py` only knows the 5-column API table, so it reads the MCP Servers table's `Transport` column as `HTTPS` and its `Install` column as `CORS`. Every row in that table already fails for that reason, including the three that are on master today.

Running the validator locally:

| README | errors |
| --- | --- |
| `master` (f4e3de1), unchanged | 593 |
| this branch | 595 |

The two added errors on my row are the same two the existing `Filesystem` row produces:

```
(L068) `stdio` is not a valid HTTPS option          <- Filesystem, already on master
(L068) – is not a valid CORS option                 <- Filesystem, already on master
(L069) `stdio` is not a valid HTTPS option          <- this PR
(L069) [Glama](...) is not a valid CORS option      <- this PR
```

So the entry matches the format CONTRIBUTING.md documents for MCP servers and matches the rows already in the table; the validator just has no MCP awareness yet. Every other open PR with checks reported is red on this same job.

Happy to open a separate PR teaching `format.py` to skip or separately validate the MCP Servers table if that would be useful — kept out of this one to respect the one-link-per-PR rule.

l3546:

✅ Instant delivery
✅ Secure & stable connection

🛒 Shop now at: tda938fe.top

